PR #222, opened 2021-07-03
fix bug: all records in app table may be overwritten during deploy under some conditions

CecilHarvey:
eg. exceptions occured, assert failed, might be other conditions as well.

Fixes Issue #218.


wolfboys:
good job!✌️

PR #270, opened 2021-07-29
Import scalatest and add unit test for MavenTool

Al-assad:

valid commit: from f179540 to 91c9acb


wolfboys (commented):
optimize the code to make the code more concise and gracefully

PR #271, opened 2021-07-30
build fat-jar before submitting flink job on k8s-session mode

Al-assad:


wolfboys (commented):
LGTM

PR #290, opened 2021-08-16
streamx-console: k8s adaptation bug fix

Al-assad:

* Missing clusterId  when submit flink k8s-job
* Application name failed to determine for k8s execution
* Force the creation of temporary processing of the LFS workspace at boot time


wolfboys (commented):
LGTM

PR #353, opened 2021-10-13
flink 1.14 support

wolfboys:
## Feature Request

1. flink 1.14 support
2. flink multi version runtime dependency isolation
3. The task can specify the version of the flink to run

----
1. flink 1.14 适配支持
2. 多版本flink运行时依赖隔离,更好的自适应on yarn和 on k8s
3. 任务可以自由选择flink版本运行


WilsonWangCS:
When I restarted streamx, streamx re-created monitoring for k8s application mode applications. If the task is not monitored, it will report an error, causing streamx to fail to start.

PR #363, opened 2021-10-18
Separate docker image pull and build process

Al-assad:

When an incorrect flink-base-image was specified to submit flink-job on flink-k8s application mode, the associated error log is vague and does not provide sufficient error information.

### What is changed ?
Separate the docker image pull and build process on DockerTool.


wolfboys:
LGTM.

PR #385, opened 2021-10-26
Temporary building resources recycling for streamx-packer module

Al-assad:

Problem Summary:
`streamx-packer` does not implement thorough temporary file recycling when building fat-jar, and more and more temporary build files will be accumulated under `app_workspace/workspace` over time.

How it Works:

The gc task is a timed deletion based on file pathname matching, which  is executed once every 6 hours, and the file retention time is 120 hours by default (the user can customize this expiration time). Delayed expiration makes it easier for users to debug the use of these build resources.

### Release Expectation: 
This feature is expected to be released in version **1.2.1**


wolfboys (commented):
LGTM. 👍👍👍

Al-assad:
@wolfboys Got it, I'll make changes later based on your suggestions

PR #398, opened 2021-11-02
Fix the bug cause by no strictly limited

WilsonWangCS:
PR Title Format:
* 1. The docker tag name should be formatted to lowerCase. It's ordered by docker build command. Otherwise，would  build image failed.
* 2. The K8sFlinkChangeEventListener.persistentK8sFlinkJobStatusChange（getOne）which change the task status will take Exception  while Two or more （same clusterId and same k8s_namespace on k8s application mode）.PerHaps,It's better to limit clusterId's Uniqueness on same Namespace by front end when we input form.


streamx-plugin/streamx-flink-packer/src/main/scala/com/streamxhub/streamx/flink/packer/docker/DockerTool.scala
```diff
@@ -44,7 +44,7 @@ object DockerTool extends Logger {
     }
     // generate dockerfile
     val dockerfile = dockerFileTemplate.writeDockerfile(projectBaseDir)
-    val tagName = compileTag(expectImageTag, authConf.registerAddress)
+    val tagName = compileTag(expectImageTag, authConf.registerAddress).toLowerCase
```
Zclhlmgqzc, line 47: Maybe it will be better
https://github.com/streamxhub/streamx/blob/6aa8cbe47ba8728267b73640453ffb2f32f8a55e/streamx-plugin/streamx-flink-packer/src/main/scala/com/streamxhub/streamx/flink/packer/docker/DockerTool.scala#L114
```scala
  def formatTag(tag: String, registerAddress: String): String = {
    if (registerAddress.nonEmpty && !tag.startsWith(registerAddress)) {
      s"$registerAddress${if (tag.contains("/")) "/" else "/library/"}$tag".toLowerCase
    } else tag.toLowerCase
  }
```

PR #412, opened 2021-11-09
Fix some bugs in special scene

WilsonWangCS:
* Restrict the DockerImage toLowerCase;
* Expand try...catch in task start, to resolve no-catch Exception;
* Change Transactional  level to REQUIRES_NEW, avoid too many backup files or rollback in FileSystem;
* Fix the bug of getOne caused by more than one same clusterId under same nameSpace;
* add new method 'download' , copy from Hdfs and LoalFs to LoalFs;


wolfboys:
LGTM

PR #448, opened 2021-11-19
Keep ASL-2.0 license from Flink source

Al-assad:


CalvinKirs (changes requested):
Good Job!
You also need to add (usually append, but you don't seem to have a NOTICE file) his NOTICE file is in your project root directory.

Al-assad:
> Good Job!
> You also need to add (usually append, but you don't seem to have a NOTICE file) his NOTICE file is in your project root directory.

Thanks for your suggestion, I will add this section later

Al-assad:
PTAL again  @CalvinKirs

NOTICE
```diff
@@ -0,0 +1,158 @@
...
+these subcomponents is subject to the terms and conditions of the
+subcomponent's license, as noted in the license.txt file.
+
+
+Apache Maven
+Copyright 2001-2019 The Apache Software Foundation
+
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
+
```
CalvinKirs, line 95: Why add this?

I am not familiar with the project itself, I see you have added a lot of things, I am not sure if these things are necessary.

CalvinKirs: It is important to keep NOTICE as brief and simple as possible, as each addition places a burden on downstream consumers.

Al-assad: Streamx uses the maven shaded plugin code directly to build the flink job jar in real time, so the NOTICE of the maven project is appended here. I'm not quite sure if I should do that, What's your opinion?

CalvinKirs: Hi, then there is no need to declare here. The NOTICE here only needs to indicate that you have made changes based on this. (FYI: For example, if you copy the Flink code in your code, it may be based on their modification, or it may be just a copy. It is usually recommended not to do this. If they provide related jars, it is best to quote directly. The reason is that I Mentioned above,)
BTW, if there are binary dependencies, you also need to create a folder to store the LICENSE and NOTICE of these dependencies.

CalvinKirs: FYI: https://infra.apache.org/licensing-howto.html

Al-assad: Thanks a bunch 🙏

PR #463, opened 2021-11-23
fix （hadoop-user-name not work #449）

monrg:
fix  hadoop-user-name not work #449

What's Changed:
add override  hadoop-user-name  config


wolfboys (commented):
LGTM.

PR #473, opened 2021-11-27
fix bug: wrong jar file chosen when building

gitfortian:
https://github.com/streamxhub/streamx/issues/471
Issue Number: close #xxx 

Problem Summary:
when building a project,it did't chose the biggest jar
change the jar.getTotalSpace to file.length
Proposal: [xxx](url) 

What's Changed:

How it Works:


wolfboys (commented):
LGTM.

PR #511, opened 2021-12-10
Refer to the list request and fix the duration value

xinzhuxiansheng:

Issue Number: close #xxx 

Problem Summary:

Proposal: [xxx](url) 

What's Changed:

How it Works:

Reference list request, fix the duration value


wolfboys (commented):
LGTM.

PR #516, opened 2021-12-13
fix the endTime value

xinzhuxiansheng:

Issue Number: https://github.com/streamxhub/streamx/issues/500

Problem Summary:

Proposal: [xxx](url) 

What's Changed:

How it Works:

fix the endTime value


wolfboys:
LGTM.

PR #555, opened 2021-12-24
Fix the wrong button icon

Al-assad:

The Cancel Task button icon is displayed incorrectly, as follows.
iShot2021-12-24 14 48 44
It should be modified to be:
iShot2021-12-24 14 51 13


wolfboys (commented):
LGTM.

PR #683, opened 2022-01-30
optimization redis sink

monrg:
optimization redis sink : add JedisSentinelPool connect, add sink mode
test hase source sink


wolfboys:
LGTM.

PR #706, opened 2022-02-13
[improve] job submission parameter optimization

wolfboys:
[improve] job submission parameter optimization

Issue Number: close #xxx 

Problem Summary:

Proposal: [xxx](url) 

What's Changed:

How it Works:


wolfboys:
@Al-assad  @czy006 PTAL

Al-assad (commented):
@wolfboys  I think the set method signature should not be the same as the original flink Configuration#set, because there is a possibility that the caller forgets to import the implicit class and uses the native flink configuration method actually.

wolfboys:
> @wolfboys I think the set method signature should not be the same as the original flink Configuration#set, because there is a possibility that the caller forgets to import the implicit class and uses the native flink configuration method actually.

safeSet implicit method ?

Al-assad (approved):
@wolfboys I'm pretty sorry I misread the method signature of config set🥲 , LGTM.

wolfboys:
> @wolfboys I'm pretty sorry I misread the method signature of config set🥲 , LGTM.

😄

PR #832, opened 2022-03-19
Improve start log

daixiaoyu:

Problem Summary:

If the application fails to start, users can choose to go to the details page

What's Changed:
Change the file   streamx-console-webapp/src/views/flink/app/View.vue

Modify the handleStartOk method to give the user the option to jump to details

How it Works:
Currently, application failure notification is not very friendly. We need to help users go to the log page


wolfboys:
LGTM

PR #848, opened 2022-03-23
Fix document path error

khotyn:
The origin path result in a 404 page:

image


wolfboys:
please visit: https://www.streamxhub.com/zh-CN/

wolfboys (approved):
LGTM.
thank you very much for your contribution and we are look forward to your next contribution

PR #892, opened 2022-04-01
optimize connector struct & streamx-flink-connector support scala and java api

monrg:


wolfboys:
LGTM. we can discussion based on this adjustment

PR #975, opened 2022-04-23
[improve] optimize es configuration load logic

monrg:
optimize es configuration load logic


wolfboys:
LGTM.

PR #1011, opened 2022-05-06
[Fix] generate token bug fixed.

numbernumberone:
Problem Summary:

Proposal: [xxx](url) 

## Contribution Checklist

  - Make sure that the pull request corresponds to a [GITHUB issue](https://github.com/streamxhub/streamx/issues).

-->


wolfboys:
@lizzy-yang PTAL.

lizzy-yang:
pr #1026

lizzy-yang:
> 
This is not a bug. The user password should not be used here. The token is generated according to username, password and expireTime. If the user password is used here, the access token generated by the same user on the browser will be exactly the same, which is not allowed！

xianweiY:
> @lizzy-yang PTAL.

In fact, the authentication of accesstoken has nothing to do with the user password, because it will directly query the database information. The user password here plays a role of salt value.

xianweiY:
> @lizzy-yang PTAL.

So this is not a bug. On the contrary, it is better to take UUID, which can make accesstoken more random

PR #1023, opened 2022-05-10
[Bug] Dynamic Option not support jvm params#1022

xxyykkxx:


streamx-console/streamx-console-service/src/main/java/com/streamxhub/streamx/console/core/service/impl/ApplicationServiceImpl.java
```diff
@@ -1270,4 +1275,17 @@ private Map<String, Serializable> getFlameGraph(Application application) {
         return flameGraph;
     }
 
+    private String[] parseDynamicOptions(String dynamicParams) {
+        List<String> dynamicOptins = new ArrayList<>();
+        Pattern pattern = Pattern.compile("-Denv.java.opts\\s*=\\s*\".*\"");
```
wolfboys: The parrent here can be promoted to a static global variable, which avoids instantiating it every time the method is called

----
这里的parrent可以提升为全局变量,这样避免每次调用 parseDynamicOptions 方法时都实例化一次

streamx-console/streamx-console-service/src/main/java/com/streamxhub/streamx/console/core/service/impl/ApplicationServiceImpl.java
```diff
@@ -1109,7 +1114,7 @@ public boolean start(Application appParam, boolean auto) throws Exception {
                 throw new UnsupportedOperationException("Unsupported...");
             }
 
-            String[] dynamicOption = CommonUtils.notEmpty(application.getDynamicOptions()) ? application.getDynamicOptions().split("\\s+") : new String[0];
+            String[] dynamicOption = CommonUtils.notEmpty(application.getDynamicOptions()) ? parseDynamicOptions(application.getDynamicOptions()) : new String[0];
```
wolfboys: This place can put the logic of judging whether application.getDynamicOptions() is null into the parseDynamicOptions method, which makes it easier to call

---

这个地方可以将判断application.getDynamicOptions()是否为null的逻辑放到 parseDynamicOptions 方法中,这样调用起来更简单.

wolfboys (commented):
It looks good, the code in some places can be improved and optimized

--

看着还不错,有些地方的代码可以在改进优化一下

wolfboys:
the parseDynamicOptions method is defined in both classes. It is repetitive logic. It can be move to com.streamxhub.streamx.common.util.FlinkUtils

----

parseDynamicOptions 方法在两个类里都定义了,是重复的逻辑,在代码实现上不太好,可以改进下,可以将该方法放到 com.streamxhub.streamx.common.util.FlinkUtils 中

streamx-common/src/main/scala/com/streamxhub/streamx/common/util/FlinkUtils.scala
```diff
@@ -21,11 +21,14 @@ package com.streamxhub.streamx.common.util
...
 import org.apache.flink.runtime.state.FunctionInitializationContext
-
 import java.io.File
+import java.util.regex.{Matcher, Pattern}
+import org.apache.commons.lang3.StringUtils
+import scala.collection.mutable.ListBuffer
 
 
 object FlinkUtils {
+  private lazy val pattern = Pattern.compile("-Denv.java.opts\\s*=\\s*\".*\"")
```
wolfboys: 这个变量名其实可以命名的更好一些,见明知意, 如: ENV_JAVA_OPTS_PATTERN. 避免后续在该类下还有其他的 parrern 变量. 

  private[this] lazy val ENV_JAVA_OPTS_PARRERN = Pattern.compile("-Denv.java.opts\\s*=\\s*\".*\"")

wolfboys (approved):
LGTM. but the variable name can be improved.

streamx-common/src/main/scala/com/streamxhub/streamx/common/util/FlinkUtils.scala
```diff
@@ -39,4 +42,23 @@ object FlinkUtils {
     }
   }
 
+  def parseDynamicOptions(dynamicParams: String): Array[String] = {
```
wolfboys, line 47: Maybe it's better?

```scala
def parseDynamicOptions(dynamicParams: String): Array[String] = {
    if (StringUtils.isBlank(dynamicParams)) Array.empty[String]; else {
      val matcher: Matcher = ENV_JAVA_OPTS_PATTERN.matcher(dynamicParams)
      val buffer = ListBuffer[String]()
      val parsedParams = if (!matcher.find) dynamicParams.trim; else {
        buffer += matcher.group.replace("\"", "")
        dynamicParams.replace(matcher.group, "").trim
      }
      buffer.++=(parsedParams.split("\\s+")).toArray
    }
  }
```

wolfboys (commented):
LGTM

PR #1047, opened 2022-05-22
[Feat]flink sql 更多类型支持

ghost:
1. 支持 create catalog
2. 支持 create function


wolfboys:
Great features, you can add some test cases, which is better.

Issue #1060, opened 2022-06-02
[Bug] find jar in error path


ChenlingXie:
### Search before asking

- [X] I had searched in the [issues](https://github.com/streamxhub/streamx/issues?q=is%3Aissue+label%3A%22bug%22) and found no similar issues.

### What happened

When building the project, the log shows that the build was successful, but failed on the ui. 
The log shows that the jar could not be found.
The suspicion is looking for the tar/jar by the folder name 'target', not by the maven folder structure.

> can't find tar.gz or jar in /opt/streamx_workspace/project/flink/hlx_bigdata_root-develop/hlx_bigdata_flink/src/main/scala/com/cn/hailanxin/flink/project/shoreline/target

### StreamX Version

1.2.3

### Flink Version

1.13.0

### Scala Version of Flink

2.11

### Error Exception

```log
2022-06-02 11:09:42.972 StreamX [streamx-build-executor-19] ERROR c.s.s.c.c.s.impl.ProjectServiceImpl:206 - deploy error, project name: flink, detail: [StreamX] can't find tar.gz or jar in /opt/streamx_workspace/project/flink/hlx_bigdata_root-develop/hlx_bigdata_flink/src/main/scala/com/cn/hailanxin/flink/project/shoreline/target
```

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wolfboys:
fixed,

PR #1064, opened 2022-06-02
[Bug] find jar in error path #1060

ChenlingXie:
find jar in all path ,all path don`t have jar ,throw exception.

Traverse all target folders, and all target folders have no jar to throw an exception.


wolfboys:
LGTM

Issue #1061, opened 2022-06-02
[Bug] Taskmanager memory usage statistics calculation error on yarn mode


ChenlingXie:
### Search before asking

- [X] I had searched in the [issues](https://github.com/streamxhub/streamx/issues?q=is%3Aissue+label%3A%22bug%22) and found no similar issues.

### What happened

Memory usage statistics calculation error, the number of taskmanger is not calculated

### StreamX Version

1.2.3

### Flink Version

1.13.1

### Scala Version of Flink

2.11

### Error Exception

```log
No Exception.
```

### Screenshots

![image](https://user-images.githubusercontent.com/11712168/171597293-36c8ba36-ce56-4fa1-ba82-f2cd5915d51a.png)

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

chuangehh:
@wolfboys 
assign me，I have a similar problem

分配给我吧，我遇到了类似的问题

PR #1065, opened 2022-06-02
[Bug] Taskmanager memory usage statistics calculation error on yarn mode #1061

ChenlingXie:
fixed error ,calculate memory use taskmanager number

image

image


wolfboys (approved):
LGTM

Issue #1074, opened 2022-06-07
[Bug] when I changed flinksql and dependency , app should not build


zhuzhihao94:
### Search before asking

- [X] I had searched in the [issues](https://github.com/streamxhub/streamx/issues?q=is%3Aissue+label%3A%22bug%22) and found no similar issues.

### What happened

when I changed flinksql and dependency , app should not build

### StreamX Version

1.2.3

### Java Version

jdk 1.8

### Flink Version

1.14.4

### Scala Version of Flink

2.12

### Error Exception

```log
null
```

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #1077, opened 2022-06-07
should build app when all changed

zhuzhihao94:
should build app when all changed #1074


wolfboys:
LGTM. thx, looking forward to your next contribution.😁

PR #1087, opened 2022-06-10
print log when update setting failed

zhuzhihao94:
1.when i update Docker Register Password on web，not get error，but setting value is old in mysql， i do not get any log in streamx log file
2.debug find update failed due to password is too long，our company requires long pwd over 255 char


wolfboys:
LGTM

PR #1113, opened 2022-06-19
Support for helm build streamx on k8s

MonsterChenzhuo:
1.Fixed docker-related bugs
2.add helm install


wolfboys:
@czy006 PTAL

czy006:
I hava some question for this mr

- helm is support spring config to yaml at next helm version ?
- For helm version and chart version, can you elaborate on the difference between them in the remarks? This parameter may be confusing for users

@MonsterChenzhuo

PR #1181, opened 2022-07-06
add ldap login

MonsterChenzhuo:
What problem to solve?
In our daily office, we will encounter a variety of office systems, imagine aha, if each system needs to maintain a set of their own user system, as a wage earner we need to remember multiple user names and passwords. If we forget a password from time to time, we would be pacing back and forth on the verge of collapse. LDAP is designed to solve this problem by using one user system for the whole company and then making each system compatible with that user system.


wolfboys:
LGTM.

PR #1213, opened 2022-07-11
add ingress configuration

MonsterChenzhuo:
Problem solved：In most cases enterprise k8s clusters/servers, are firewalled and will only expose port 80 of the master node, the rest of the ports need to be accessed through nginx domain hops, which results in the flink task web interface not being available for streamx link

Add two ways to increase the ingres configuration.
1. By default, the user can just specify the domain name, and the access address is by default: domain name+namespace+clusterId
2. You are very familiar with ingress and know what you want to do, you can submit a yaml of ingress to do the configuration of ingress


wolfboys:
LGTM

chenshaojin:
I reviewed the source code and the second approach does not have a frontend interface, right? The only way is through swagger or postman. @MonsterChenzhuo

Issue #1072, opened 2022-06-06
[Bug] CICD maven module should not contain version number


ChenlingXie:
### Search before asking

- [X] I had searched in the [issues](https://github.com/streamxhub/streamx/issues?q=is%3Aissue+label%3A%22bug%22) and found no similar issues.

### What happened

After the maven module version number is changed, the module and program jar cannot be found.

### StreamX Version

1.2.3

### Flink Version

1.13.1

### Scala Version of Flink

2.11

### Error Exception

```log
No
```

### Screenshots

<img width="848" alt="c6466943758ee11c2e9b1e6f64a61cd" src="https://user-images.githubusercontent.com/11712168/172161065-f3c3dcf4-c97e-4ade-9693-94a19c63ee60.png">

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #1223, opened 2022-07-14
[BUG] module name should not contain version number. (#1072)

ChenlingXie:
module name should not contain version number.
Now modeule artifactId is the module name.
image


MonsterChenzhuo:
@ChenlingXie Sorry I don't quite understand the meaning of this pr, can you elaborate on it?

PR #1236, opened 2022-07-17
[Improve] Update upgrade.sql to save history email alert conf

WeiJinglun:
Update upgrade.sql to save history email alert conf


wolfboys:
LGTM

Issue #1204, opened 2022-07-11
12. [Bug] UDF is uploaded in the form of jar package upload. After modifying the UDF logic, upload again. When the jar package name is unchanged, it will not be rebuilt image


wolfboys:

monrg:
I will fix it .

PR #1242, opened 2022-07-18
[BUG] UDF update isn't new name  it can't work(#1204)

monrg:
<!--

Thank you for contributing to StreamX! Please make sure that your code changes
are covered with tests. And in case of new features or big changes
remember to adjust the documentation.

Issue Number: 1204

Problem Summary:

UDF is uploaded in the form of jar package upload. After modifying the UDF logic, upload again. When the jar package name is unchanged, it will not be rebuilt image

add uploadm file md5 value ,use md5 equils file if is new .


streamx-console/streamx-console-service/src/main/java/com/streamxhub/streamx/console/core/entity/Application.java
```diff
@@ -793,6 +798,11 @@ public static Dependency toDependency(String dependency) {
             return new Dependency();
         }
 
+        public void extractJarMd5() {
```
wolfboys, line 801: Compare whether the two files are the same. Take a look at the eaquls method in FileUtils. Using MD5 will first read all bytes into memory and calculate MD5. This method is not efficient, so it is not recommended

PR #1263, opened 2022-07-21
[improve] add AlertException

WeiJinglun:
add AlerException.

Feedback the detailed abnormal content of the front end. 

[Fix] When deleting an alert configuration, there is no check for application bindings.


wolfboys:
LGTM

PR #1273, opened 2022-07-22
[BUG] fix same name udf updated, it can't work(#1204)

monrg:
#1204 

Problem Summary:
fix   DF is uploaded in the form of jar package upload.  can't work.

cahnge Compare file name  to  compare file content


wolfboys:
LGTM

Issue #1208, opened 2022-07-11
16. [Bug] The flink-conf that synchronizes flink-home does not take effect


wolfboys:

Su1024:
i will fix this issues

Su1024:
#1275  can fix this issues

PR #1275, opened 2022-07-22
[BUG] Fix the problem that flink-conf synchronization does not take effect (#1208):

Su1024:
Fix the problem that flink-conf synchronization does not take effect (#1208):


wolfboys (approved):
LGTM.

PR #1283, opened 2022-07-27
[BUG] Fix the job status bug And add the getSqlClientJar exception optimization

youyangkou:

Fix the status bug


wolfboys:
LGTM.

PR #1306, opened 2022-08-01
 [Bug] Flink Job Failure Cannot Trigger Alarm

xujiangfeng001:

Issue Number: https://github.com/streamxhub/streamx/issues/1298

Problem Summary: The main reason for this problem is that when the alarm system judges the status of the Flink job, it uses the Yan state instead of the finalstatus, which makes the Flink job unable to send alarm information when it fails.

Proposal: https://github.com/streamxhub/streamx/pull/1306

## Contribution Checklist

  - Make sure that the pull request corresponds to a [GITHUB issue](https://github.com/streamxhub/streamx/issues/1298).

-->

This PR mainly solves two problems:

1. Solve the problem that the alarm may not be triggered after Flink job fails
2. Add Flink job to trigger the alarm function when the job is cancelled
On the second point, it is a place that needs to be discussed. Is it necessary to trigger the alarm function when the Flink job is cancelled?
I hope all developers can provide valuable comments on the second point above.


youyangkou:
I don't agree with this Pr. Task creators and developers are likely to start and stop frequently during debugging, and sending frequent email alarms at this time may not be good for the user experience.But The task creator's task has been stopped by someone else, and this can warn the police.

-1

streamx-console/streamx-console-service/src/main/java/com/streamxhub/streamx/console/core/task/FlinkTrackingTask.java
```diff
@@ -510,7 +510,7 @@ private void getFromYarnRestApi(Application application, StopFrom stopFrom) thro
                     cleanOptioning(optionState, application.getId());
                     this.persistentAndClean(application);
 
-                    if (flinkAppState.equals(FlinkAppState.FAILED) || flinkAppState.equals(FlinkAppState.LOST)) {
+                    if (flinkAppState.equals(FlinkAppState.FAILED) || flinkAppState.equals(FlinkAppState.LOST) || flinkAppState.equals(FlinkAppState.CANCELED)) {
```
wolfboys: This will cause all normal cancelled jobs to send alarm notifications ,We can discuss about this

WeiJinglun: Tasks are stopped by user clicks and can be divided into two categories:
1. The action is triggered by the task creator. Then, it is not recommended to send a message notification, such operations are performed independently by the task owner.
2. By a non-task creator, such as the admin user or an administrator in the user group. Then this type of operation can send a message notification, the purpose of which notifies the task creator

xujiangfeng001: > Tasks are stopped by user clicks and can be divided into two categories:
> 
> 1. The action is triggered by the task creator. Then, it is not recommended to send a message notification, such operations are performed independently by the task owner.
> 2. By a non-task creator, such as the admin user or an administrator in the user group. Then this type of operation can send a message notification, the purpose of which notifies the task creator

I agree with this approach. In addition, I think the following situations also need to trigger the alarm mechanism:
The action of job cancellation is initiated by an external system (such as Flink Web UI, command line, etc.). Because the job cancellation action initiated by the external system has broken away from the original user system of streamx, I think it should also be Treat it as an exception.

xujiangfeng001:
> I don't agree with this Pr. Task creators and developers are likely to start and stop frequently during debugging, and sending frequent email alarms at this time may not be good for the user experience.But The task creator's task has been stopped by someone else, and this can warn the police.
> 
> -1

Thank you for your comments. Please see below for details

streamx-console/streamx-console-service/src/main/java/com/streamxhub/streamx/console/core/task/FlinkTrackingTask.java
MonsterChenzhuo: +1，Your pr I understand is an alarming strategy, suggest adding an optional configuration item in the UI layer and the back-end system layer

xujiangfeng001:
After unanimous discussion, this PR is only merged https://github.com/streamxhub/streamx/issues/1298 Question.
For the improvement of the alarm mechanism, put https://github.com/streamxhub/streamx/issues/1313 Submit

wolfboys:
LGTM

PR #1307, opened 2022-08-01
remove 'mast have insert statement' limit

ziqiang-wang:
We don't mast write insert statement in flink sql job.
Each Flink SQL statement is executed immediately and individually, with corresponding effects.

### DIFF
--- a/streamx-flink/streamx-flink-shims/streamx-flink-shims-base/src/main/scala/com/streamxhub/streamx/flink/core/FlinkSqlExecutor.scala
+++ b/streamx-flink/streamx-flink-shims/streamx-flink-shims-base/src/main/scala/com/streamxhub/streamx/flink/core/FlinkSqlExecutor.scala
@@ -145,6 +145,8 @@ object FlinkSqlExecutor extends Logger {
         case SELECT =>
           logError("StreamX dose not support 'SELECT' statement now!")
           throw new RuntimeException("StreamX dose not support 'select' statement now!")
+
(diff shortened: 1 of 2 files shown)


MonsterChenzhuo:
@ziqiang-wang 
Are you saying that pure create statements can be committed
select statements can also be committed

ziqiang-wang:
1. The results of the SELECT statement cannot be collected locally, so the platform does not support the SELECT statement yet.
2. Each statement is executed immediately. For example, if the 'Create Hive Catalog' statement is written first and then the 'Create Table' statement is written, the metadata of the table will written to the Hive metadata.

ziqiang-wang:
@MonsterChenzhuo 
The results of the SELECT statement cannot be collected locally, so the platform does not support the SELECT statement yet.
Each statement is executed immediately. For example, if the 'Create Hive Catalog' statement is written first and then the 'Create Table' statement is written, the metadata of the table will written to the Hive metadata.

MonsterChenzhuo:
@ziqiang-wang He solved what pain point problem, I did not he understand his applicable scenario

wolfboys:
The current version requires an insert operation so that the job can run

ziqiang-wang:
This change was not necessary, I just discovered this feature.

For example, I created the Hive Catalog, and then only run some `create table` statement. Then other tasks do not need to write `create table` statement, and directly write an `INSERT` statement to run the job, which can reduce the redundancy of create table statements in job development.

PR #1333, opened 2022-08-09
[Feature]Support for creating project groups

MonsterChenzhuo:
Support the creation of project teams to coordinate tasks by the project team, with the effect of task isolation


lvshaokang:
Could you fix the license header follow the class?
[StreamXConsole.java](https://github.com/streamxhub/streamx/blob/dev/streamx-console/streamx-console-service/src/main/java/com/streamxhub/streamx/console/StreamXConsole.java)

Issue #1325, opened 2022-08-08
Polish maven wrapper and bump maven version to 3.8.6 for better performance and arm support.


GOODBOY008:
### Search before asking

- [X] I had searched in the [issues](https://github.com/streamxhub/streamx/issues?q=is%3Aissue+label%3A%22bug%22) and found no similar issues.

### What happened

Remove jar and class file in dictionary `.mvn/wrapper`,add `wrapperUrl` config to `maven-wrapper.properties`.
Upgrade maven-wrapper to `3.1.0`, maven to `3.8.6`(at least `3.8.4`) for performance improvements and ARM more native support.

### StreamX Version

none 

### Java Version

none 

### Flink Version

none 

### Scala Version of Flink

none 

### Error Exception

```log
none
```

### Screenshots

none 

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

GOODBOY008:
https://github.com/apache/incubator-streampark/pull/1334 Merged.

PR #1334, opened 2022-08-09
[Improve] Polish maven wrapper and bump maven version to 3.8.4 for better performance and arm support.

GOODBOY008:
Link issue #1325 
[Improve] Polish maven wrapper and bump maven version to 3.8.4 for better performance and arm support.


wolfboys:
LGTM

PR #1366, opened 2022-08-12
[polish] module streamx-console-service's log error info to file instead of print error stack.

GOODBOY008:
[polish] module streamx-console-service's log error info to file instead of print error stack.


wolfboys:
LGTM

PR #1367, opened 2022-08-12
[Feat] init storage service for jar upload/download to support flink on native k8s mode

legendtkl:
init module `streamx-storage-service`.

We will use this module to support flink job on native k8s mode in uploading and downloading jar files to/from remote storage, such as OSS/HDFS, etc.


wolfboys:
LGTM

PR #1402, opened 2022-08-18
[polish] set spark moudle default scala version to 2.12.

GOODBOY008:
[polish] set default scala version to 2.12.


wolfboys:
LGTM

PR #1550, opened 2022-09-08
improve: more notifications settings in .asf.yaml

tisonkun:
Dispatch notification in fine-grained.

Also, we may need one more commit to trigger the watcher for applying this change. IIRC the initialization happened before repo transformation.

And default branch `dev` may not be recognized by the automation. Will file an issue on INFRA to fix it if it's the case.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -39,7 +39,7 @@ github:
 
 notifications:
   commits:      commits@streampark.apache.org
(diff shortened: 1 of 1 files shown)


wolfboys:
LGTM. 
This file was added by me with reference to other apache projects. thanks for your guidance

wolfboys (approved):
LGTM

PR #1559, opened 2022-09-09
[Bug]Fixed some job properties are not replicated bug When copy a job

youyangkou:
Fixed some job properties are not replicated bug When copy a job
#### Problem Summary:
When copying a job, some job properties are not replicated and there are some bug.
For example：
tags and ModifiedTime is empty
Versionid directly wrote as 10000L and there will be flink-home version error when use postgres as database

Proposal: [xxx](url) 

## Contribution Checklist

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/ApplicationServiceImpl.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/ApplicationServiceImpl.java
@@ -623,7 +623,7 @@ public class ApplicationServiceImpl extends ServiceImpl<ApplicationMapper, Appli
         newApp.setClusterId(oldApp.getExecutionModeEnum() == ExecutionMode.KUBERNETES_NATIVE_SESSION ? oldApp.getClusterId() : jobName);
         args = args != null && !"".equals(args) ? args : oldApp.getArgs();
         newApp.setArgs(args);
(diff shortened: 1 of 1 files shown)


youyangkou:
@wolfboys PTAL

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/ApplicationServiceImpl.java
```diff
@@ -623,7 +623,7 @@ public Long copy(Application appParam) {
         newApp.setClusterId(oldApp.getExecutionModeEnum() == ExecutionMode.KUBERNETES_NATIVE_SESSION ? oldApp.getClusterId() : jobName);
         args = args != null && !"".equals(args) ? args : oldApp.getArgs();
         newApp.setArgs(args);
-        newApp.setVersionId(100000L);
+        newApp.setVersionId(oldApp.getVersionId());
```
wolfboys, line 626: no. create app from copy, versionId cannot be the same as the original.

wolfboys: > no. create app from copy, versionId cannot be the same as the original.

sorry, here is the version of flink bound. It's ok.

wolfboys (approved):
LGTM

PR #1579, opened 2022-09-12
[Feature] implement Oss Storage Service

legendtkl:
### Purpose of this pull request
We will support several remote storage services in StreamPark, and these storage services will be used in Flink/Spark on k8s for pod template as an artifacts upload/fetch.

This PR implement Aliyun OSS supports.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -99,7 +99,7 @@
         <hive.version>2.3.4</hive.version>
         <hadoop.version>2.10.2</hadoop.version>
         <hbase.version>2.1.10</hbase.version>
-        <junit.version>5.6.3</junit.version>
+        <jupiter.version>5.6.3</jupiter.version>
         <zkclient.version>0.11</zkclient.version>
         <curator.version>4.2.0</curator.version>
         <redis.version>3.3.0</redis.version>
@@ -128,6 +128,9 @@
         <streampark.shaded.package>org.apache.streampark.shaded</streampark.shaded.package>
         <flink.table.uber.artifact.id>flink-table-uber_${scala.binary.version}</flink.table.uber.artifact.id>
         <httpclient.version>4.5.13</httpclient.version>
+        <lombok.version>1.18.24</lombok.version>
+        <aliyun.oss.version>3.15.0</aliyun.oss.version>
+        <junit.version>4.13.2</junit.version>
 
         <PermGen>64m</PermGen>
         <MaxPermGen>512m</MaxPermGen>
@@ -534,6 +537,24 @@
                 <version>${log4j.version}</version>
             </dependency>
             <!--log4j end-->
+
+            <dependency>
+                <groupId>org.projectlombok</groupId>
+                <artifactId>lombok</artifactId>
+                <version>${lombok.version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>com.aliyun.oss</groupId>
+                <artifactId>aliyun-sdk-oss</artifactId>
+                <version>${aliyun.oss.version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>junit</groupId>
+                <artifactId>junit</artifactId>
+                <version>${junit.version}</version>
+            </dependency>
 
         </dependencies>
 
--- a/pom.xml
+++ b/pom.xml
@@ -129,7 +129,6 @@
         <flink.table.uber.artifact.id>flink-table-uber_${scala.binary.version}</flink.table.uber.artifact.id>
         <httpclient.version>4.5.13</httpclient.version>
         <lombok.version>1.18.24</lombok.version>
-        <aliyun.oss.version>3.15.0</aliyun.oss.version>
         <junit.version>4.13.2</junit.version>
 
         <PermGen>64m</PermGen>
--- a/pom.xml
+++ b/pom.xml
@@ -544,12 +544,6 @@
             </dependency>
 
             <dependency>
-                <groupId>com.aliyun.oss</groupId>
-                <artifactId>aliyun-sdk-oss</artifactId>
-                <version>${aliyun.oss.version}</version>
-            </dependency>
-
-            <dependency>
                 <groupId>junit</groupId>
                 <artifactId>junit</artifactId>
                 <version>${junit.version}</version>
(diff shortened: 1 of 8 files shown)


pom.xml
```diff
@@ -535,6 +538,24 @@
...
+                <groupId>org.projectlombok</groupId>
+                <artifactId>lombok</artifactId>
+                <version>${lombok.version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>com.aliyun.oss</groupId>
+                <artifactId>aliyun-sdk-oss</artifactId>
+                <version>${aliyun.oss.version}</version>
+            </dependency>
```
wolfboys: The root pom declare global general dependencies, such as oss, s3, etc., which are the dependencies that only need to be used by streampark-storeage module. so, it is recommended to define it in streampark-storeage module

streampark-storage/src/main/java/org/apache/streampark/storage/oss/OssStorageService.java
```diff
@@ -0,0 +1,92 @@
...
+import com.aliyun.oss.OSSClientBuilder;
+import com.aliyun.oss.OSSException;
+import com.aliyun.oss.model.GetObjectRequest;
+import com.aliyun.oss.model.PutObjectRequest;
+import com.google.common.annotations.VisibleForTesting;
+import lombok.extern.slf4j.Slf4j;
+
+import java.io.File;
+
+@Slf4j
```
1996fanrui, line 33: The `@Slf4j` isn't used, right?

pom.xml
legendtkl: addressed.

streampark-storage/src/main/java/org/apache/streampark/storage/StorageService.java
```diff
@@ -17,8 +17,11 @@
 
 package org.apache.streampark.storage;
 
+/**
+ * StorageService will be used as artifacts fetcher in pod template, so it should rely on other modules.
+ */
 public interface StorageService {
```
1996fanrui, line 23: StorageService just be used as artifacts fetcher in pod template, right?

If yes, the class name is unreasonable. The StorageService looks like a generic storage service.

1996fanrui (changes requested):
@legendtkl thanks for your contribution, I left some comments, please take a look in your free time.

pom.xml
1996fanrui: I see you just move properties to `streampark-storage` module.

As I understand, @wolfboys  means move these dependencies to `streampark-storage` module, not just move properties. Root pom shouldn't add these dependencies, they should be added in `streampark-storage` module.

streampark-storage/src/main/java/org/apache/streampark/storage/oss/OssStorageService.java
legendtkl: @1996fanrui  Good catch. I will update the code.

streampark-storage/src/main/java/org/apache/streampark/storage/StorageService.java
legendtkl: > StorageService just be used as artifacts fetcher in pod template, right?
> 
> If yes, the class name is unreasonable. The StorageService looks like a generic storage service.

@1996fanrui  No, this will be used to uploading job-related resources to remote storage system.

1996fanrui: Thanks for you quick reply.

If no, I have some questions:

1. why did you add the comments for  `StorageService`? 

> StorageService will be used as artifacts fetcher in pod template, so it should rely on other modules.

2. Whether the localFilePath parameter of 2 methods are reasonable? 

- I worry that the current interface definition only solves the current needs and will not be scalable in the future. 
- For example, will the distributed file system be supported in the future? Or just download the data into memory.
- Would it be more general to replace `String localFilePath` with `OutputStream/InputStream`?

3. Whether the objectPath of 2 methods are reasonable? 

- As I understand, ObjectStorage is just one type of StorageService, it shouldn't always be `String objectPath`, right?

Please correct me if there are any mistakes.

1996fanrui:
Hi @legendtkl , I still have some questions about these 2 comments, could you un-resolve them? thanks a lot.

https://github.com/apache/incubator-streampark/pull/1579#discussion_r968448518

https://github.com/apache/incubator-streampark/pull/1579#discussion_r968437610

wolfboys:
Hi @legendtkl , root pom just defined global general dependencies, e.g: scala, jackson, log4j,  many modules used. aliyun-sdk-oss only used in streampark-storeage module, so, aliyun-sdk-oss  dependency in the root POM should be deleted. and defined in streampark-storeage module.
e.g: streampark-console use a lot of spring dependencies, but they are not defined in the root POM, defined in streampark-consoel module. 


-----
root pom只用来定义全局通用的依赖, 例如: scala,jackson,log4j等, 很多模块都使用到了, 但是aliyun-sdk-oss这个只在streampark-storeage中被使用,是某个模块特有的依赖,不应该定义到root pom中,应该从root pom中删除,定义到具体的模块中, 如: 在streampark-console中,大量使用了spring的依赖,但是他们并没有被定义到root pom中,只是定义在streampark-consoel 模块中.

pom.xml
wolfboys: > I see you just move properties to `streampark-storage` module.
> 
> As I understand, @wolfboys means move these dependencies to `streampark-storage` module, not just move properties. Root pom shouldn't add these dependencies, they should be added in `streampark-storage` module.

yes

legendtkl:
> Hi @legendtkl , root pom just defined global general dependencies, e.g: scala, jackson, log4j, many modules used. aliyun-sdk-oss only used in streampark-storeage module, so, aliyun-sdk-oss dependency in the root POM should be deleted. and defined in streampark-storeage module. e.g: streampark-console use a lot of spring dependencies, but they are not defined in the root POM, defined in streampark-consoel module.
> 
> root pom只用来定义全局通用的依赖, 例如: scala,jackson,log4j等, 很多模块都使用到了, 但是aliyun-sdk-oss这个只在streampark-storeage中被使用,是某个模块特有的依赖,不应该定义到root pom中,应该从root pom中删除,定义到具体的模块中, 如: 在streampark-console中,大量使用了spring的依赖,但是他们并没有被定义到root pom中,只是定义在streampark-consoel 模块中.

@wolfboys  resolved.

wolfboys (approved):
LGTM

PR #1595, opened 2022-09-13
[Improve] remove unnecessary test class to avoid thread blocking.

lvyanquan:
RefreshCacheTest will block the thread while running all test cases and will print lines as following, 
```
2022-09-13 20:24:13
2022-09-13 20:24:23
2022-09-13 20:24:33
2022-09-13 20:24:43
2022-09-13 20:24:53
2022-09-13 20:25:03
2022-09-13 20:25:13
```
and 'while(true)' loop will also block the thread,
```java
while (true) {
    System.out.println(caffeine.getIfPresent("config"));
    Thread.sleep(2000);
}
``` 
however, Interface of RefreshCache has empty body, thus this test class is unnecessary.
This pr will remove [RefreshCacheTest](https://github.com/apache/incubator-streampark/blob/dc9150941d0873a88c851e42c64b5de3e12d85ba/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/annotation/RefreshCache.java) class.


wolfboys:
I actually tested locally, both methods are as expected, please tall me why you should delete them, if have bug, you can fix it.

wolfboys (approved):
LGTM

PR #1606, opened 2022-09-14
[Hotfix] Re-arrange the test package structure of streampark-console-service.

RocMarshal:
Problem Summary:

Proposal: Only code-cleanup work. Re-arrange the test package structure of the streampark-console-service module.

## Contribution Checklist

Re-arrange the test package structure of the streampark-console-service module.


wolfboys:
Thanks for your contribution. I've seen many package name has changed, it is not under the same package name,  Can you tell me why we need to change like this?

RocMarshal:
@wolfboys 
- Some test cases about built-in classes like ```CompletableFutureTest.java```、```RegexTest```  should not be located in the repo. Because It looks more like a usage example. However, the tested classes have already be stable.
- Some test cases about third-part dependencies like ```caffeine``` are also same as mentioned above.
- The test cases classes files would be better located in the same package which contains target tested classes to test  under the corresponding test module.  In addition, other modules' test cases package structure also follow this rule

streampark-common/src/test/scala/org/apache/streampark/common/util/YarnTest.scala
```diff
@@ -0,0 +1,53 @@
...
+    val numYarnMaxVcores = HadoopUtils.yarnClient.getNodeReports(NodeState.RUNNING).stream.mapToInt(report => report.getCapability.getVirtualCores).max.orElse(0)
+    System.out.println(numYarnMaxVcores)
+  }
+
+  @Test def getURL(): Unit = {
+    /**
+     * 将hadoop的配置文件放到一个目录下,
+     * 在运行该类的时候加上jvm级别的参数(idea里的 vmOption ) -DHADOOP_CONF_DIR=${目录}
+     */
+    val url = YarnUtils.getRMWebAppURL
```
lvshaokang: @RocMarshal Thanks for you feedback. Could you handle these non-ascaii characters by remove or translate ways?

lvshaokang (changes requested):

streampark-console/streampark-console-service/src/test/java/com/github/benmanes/caffeine/cache/RefreshCacheTest.java
```diff
@@ -0,0 +1,67 @@
...
+        System.out.println(DateUtils.format(new Date(), DateUtils.fullFormat(), TimeZone.getDefault()));
+        Timer timer = new Timer();
+        timer.schedule(new TimerTask() {
+            @Override
+            public void run() {
+                System.out.println(DateUtils.format(new Date(), DateUtils.fullFormat(), TimeZone.getDefault()));
+            }
+        }, 1000 * 10, 1000 * 10);
+
+        Thread.sleep(Long.MAX_VALUE);
```
wolfboys: `Thread.sleep(Long.MAX_VALUE)` will block the thread. You can delete this testcase, see: https://github.com/apache/incubator-streampark/pull/1595

RocMarshal: @wolfboys Thx for the comment. 
Updated. PTAL.

streampark-common/src/test/scala/org/apache/streampark/common/util/YarnTest.scala
RocMarshal: @lvshaokang Thx for the comment.
Updated. PTAL.

wolfboys (dismissed):
LGTM

wolfboys (commented):
the YarnTest failed, 😞 this testcase can be deleted

wolfboys (approved):
LGTM

PR #1613, opened 2022-09-15
[build] speed up build progress.

GOODBOY008:
[build] speed up build progress.

### DIFF
--- a/streampark-common/pom.xml
+++ b/streampark-common/pom.xml
@@ -198,6 +198,33 @@
     <build>
         <plugins>
             <plugin>
+                <groupId>net.alchim31.maven</groupId>
+                <artifactId>scala-maven-plugin</artifactId>
+                <version>4.3.0</version>
+                <configuration>
+                    <addScalacArgs>-target:jvm-${project.build.jdk}</addScalacArgs>
+                    <target>${project.build.jdk}</target>
+                    <source>${project.build.jdk}</source>
+                </configuration>
+                <executions>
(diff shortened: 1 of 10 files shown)


streampark-common/pom.xml
```diff
@@ -197,6 +197,33 @@
...
+                    </execution>
+                    <execution>
+                        <id>test-compile-scala</id>
+                        <phase>test-compile</phase>
+                        <goals>
+                            <goal>testCompile</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
```
wolfboys, line 203: scala-maven-plugin is repeatedly defined in many modules. Can it be defined in the root pom using pluginManagement and referenced by other submodules?

GOODBOY008: > scala-maven-plugin is repeatedly defined in many modules. Can it be defined in the root pom using pluginManagement and referenced by other submodules?

It will make only java module build slowly.

wolfboys: > > scala-maven-plugin is repeatedly defined in many modules. Can it be defined in the root pom using pluginManagement and referenced by other submodules?
> 
> It will make only java module build slowly.

Does the scala-maven-plugin have to be fully defined in each submodule for this to work?

GOODBOY008: This pr just tuning performance by solving these issues.

wolfboys (approved):
LGTM

Issue #1561, opened 2022-09-09
[Bug]运行自定义jar包报错


wangzhen-wudangshan:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/streampark/issues?q=is%3Aissue+label%3A%22bug%22) and found no similar issues.

### What happened

运行自定义jar包报错

### StreamPark Version

streamx-1.2.3

### Java Version

jdk 1.8

### Flink Version

flink-1.13.5

### Scala Version of Flink

2.12

### Error Exception

```log
at org.apache.shiro.spring.security.interceptor.AopAllianceAnnotationsAuthorizingMethodInterceptor$1.proceed(AopAllianceAnnotationsAuthorizingMethodInterceptor.java:82)
        at org.apache.shiro.authz.aop.AuthorizingMethodInterceptor.invoke(AuthorizingMethodInterceptor.java:39)
        at org.apache.shiro.spring.security.interceptor.AopAllianceAnnotationsAuthorizingMethodInterceptor.invoke(AopAllianceAnnotationsAuthorizingMethodInterceptor.java:115)
        at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186)
        at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.proceed(CglibAopProxy.java:753)
        at org.springframework.aop.interceptor.ExposeInvocationInterceptor.invoke(ExposeInvocationInterceptor.java:97)
        at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186)
        at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.proceed(CglibAopProxy.java:753)
        at org.springframework.aop.framework.CglibAopProxy$DynamicAdvisedInterceptor.intercept(CglibAopProxy.java:698)
```

### Screenshots

![6c5ec34ed9badd2dda7f86ff4be2a52](https://user-images.githubusercontent.com/96755853/189296184-87d6997c-474d-426f-9376-87aec352913e.png)

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wangzhen-wudangshan:
[...]

PR #1624, opened 2022-09-16
[Improve] Replace the assert with checkState

### LINKED ISSUE #1623: [Improve] Replace the assert with checkState
### Search before asking

### Description

From #1561 , we found `assert` isn't convenient for troubleshooting.

I suggest that replacing the assert with checkState. If the state is false, it will throw IllegalStateException, and we can get the call stack. It can help us to troubleshooting.

### Usage Scenario

_No response_

### Related issues

_No response_

### Code of Conduct

1996fanrui:

Issue Number: close #1623 

Problem Summary:

From #1561 , we found `assert` isn't convenient for troubleshooting.

I suggest that replacing the assert with checkState. If the state is false, it will throw IllegalStateException, and we can get the call stack. It can help us to troubleshooting.

Proposal: [xxx](url) 

## Contribution Checklist

From #1561 , we found `assert` isn't convenient for troubleshooting.

I suggest that replacing the assert with checkState. If the state is false, it will throw IllegalStateException, and we can get the call stack. It can help us to troubleshooting.

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/runner/EnvInitializer.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/runner/EnvInitializer.java
@@ -96,7 +96,7 @@ public class EnvInitializer implements ApplicationRunner {
             .filter(springEnv::containsProperty)
             .forEach(key -> {
                 InternalOption config = InternalConfigHolder.getConfig(key);
-                assert config != null;
+                AssertUtils.state(config != null);
                 InternalConfigHolder.set(config, springEnv.getProperty(key, config.classType()));
             });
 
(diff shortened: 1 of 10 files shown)


streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/runner/EnvInitializer.java
```diff
@@ -96,7 +97,7 @@ private void initInternalConfig(Environment springEnv) {
             .filter(springEnv::containsProperty)
             .forEach(key -> {
                 InternalOption config = InternalConfigHolder.getConfig(key);
-                assert config != null;
+                Preconditions.checkState(config != null);
```
wolfboys: use `AssertUtils.state(...)` replace `Preconditions.checkState`, We should try to avoid use flink methods in the streampark-console, because streampark-console need to support multiple versions of flink, It is easy to cause conflicts when use flink methods

1996fanrui: Thanks for your review, updated.

wolfboys (approved):
LGTM

PR #1675, opened 2022-09-24
Revert "chore: update star history chart"

wolfboys:
Reverts apache/incubator-streampark#1672


lvshaokang (approved):
Revert, after discussion, we agree that the chart style is not compatible with the original style. cc @boojack

lvshaokang:
LGTM

PR #1725, opened 2022-09-30
[Improve]When the flink task on k8s keeps restarting, streampark won'…

MonsterChenzhuo:
## What problem does this PR solve?

When the flink task on k8s keeps restarting, streampark won't be able to do anything about it

## What is changed and how it works?

He will delete the deployment, ingress, congifmap of the task, and also save the error log for problem tracing


wolfboys:
Thanks for your contribution, I will review it later

wolfboys (approved):
LGTM

PR #1799, opened 2022-10-10
[Improve] Add file name interception

xinzhuxiansheng:
## What problem does this PR solve?

Issue Number: close #xxx 

Problem Summary:

## What is changed and how it works?

Proposal: url address 

## Contribution Checklist

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/ApplicationServiceImpl.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/ApplicationServiceImpl.java
@@ -103,6 +103,7 @@ import io.fabric8.kubernetes.client.KubernetesClientException;
 import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.flink.configuration.CheckpointingOptions;
 import org.apache.flink.configuration.RestOptions;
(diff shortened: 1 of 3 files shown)


wolfboys (approved):
LGTM

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/ApplicationServiceImpl.java
```diff
@@ -293,7 +294,8 @@ public void tailMvnDownloading(Long id) {
     @Override
     public String upload(MultipartFile file) throws ApplicationException {
         File temp = WebUtils.getAppTempDir();
-        File saveFile = new File(temp, Objects.requireNonNull(file.getOriginalFilename()));
+        String fileName = FilenameUtils.getName(Objects.requireNonNull(file.getOriginalFilename()));
+        File saveFile = new File(temp, fileName);
         // delete when exists
         if (saveFile.exists()) {
```
WillemJiang, line 300: Not sure if it can cause some side effects.

streampark-console/streampark-console-service/src/test/java/org/apache/streampark/console/core/service/ApplicationServiceTest.java
```diff
@@ -89,4 +94,15 @@ public void start() throws Exception {
         applicationService.start(application, false);
     }
 
+    @Test
+    public void uploadTest() throws Exception {
```
WillemJiang, line 98: I think we need to add another test for verifying the uploaded file is not a jar file.
It looks like we just verify the ContentType

wolfboys: > I think we need to add another test for verifying the uploaded file is not a jar file. It looks like we just verify the ContentType

Thanks very much for your review, next we will provide test cases to verify the upload file type.

PR #1848, opened 2022-10-16
fix(sec): upgrade org.apache.hadoop:hadoop-hdfs to 3.3.2

chncaption:
### What happened？
There are 1 security vulnerabilities found in org.apache.hadoop:hadoop-hdfs 2.10.2
- [MPS-2022-12474](https://www.oscs1024.com/hd/MPS-2022-12474)

### What did I do？
Upgrade org.apache.hadoop:hadoop-hdfs from 2.10.2 to 3.3.2 for vulnerability fix

### What did you expect to happen？
Ideally, no insecure libs should be used.

### How was this patch tested?
Run `mvn compile` failed locally, couldn't complete the build process.
Run `mvn clean test` failed locally, unit-test couldn't pass.

### The specification of the pull request
[PR Specification](https://www.oscs1024.com/docs/pr-specification/) from OSCS

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -96,7 +96,7 @@
         <spark.version>3.2.0</spark.version>
         <scalikejdbc.version>4.0.0</scalikejdbc.version>
         <hive.version>2.3.4</hive.version>
-        <hadoop.version>2.10.2</hadoop.version>
+        <hadoop.version>3.3.2</hadoop.version>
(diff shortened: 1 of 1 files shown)


wolfboys:
Thanks for your contribution, considering the compatibility of hadoop versions and other issues, we will not upgrade for current version, and we will discuss with the developers to see how to solve it in the next versions.

PR #1868, opened 2022-10-19
[Improve] after change team the routers need to be regenerated

### LINKED ISSUE #1867: [Improve] after change team the router need to be regenerated
### Search before asking

### Description

When switching teams, the backend to re-return the user's latest role, permissions, and router, the front-end needs to re-render the router

Currently, the backend does not return the latest router, The front-end also did not rebuild the generated router

### Usage Scenario

_No response_

### Related issues

_No response_

### Code of Conduct

wolfboys:
When switching teams, the backend to re-return the user's latest role, permissions, and router, the front-end needs to re-render the router

Currently, the backend does not return the latest router, The front-end also did not rebuild the generated router

Issue Number: close #1867 

## Verifying this change

*(or)*

*(or)*

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/service/impl/UserServiceImpl.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/service/impl/UserServiceImpl.java
@@ -21,6 +21,7 @@ import org.apache.streampark.common.util.AssertUtils;
 import org.apache.streampark.console.base.domain.RestRequest;
 import org.apache.streampark.console.base.exception.ApiAlertException;
 import org.apache.streampark.console.base.util.ShaHashUtils;
+import org.apache.streampark.console.system.authentication.JWTToken;
 import org.apache.streampark.console.system.entity.Member;
 import org.apache.streampark.console.system.entity.Menu;
 import org.apache.streampark.console.system.entity.Team;
@@ -28,6 +29,7 @@ import org.apache.streampark.console.system.entity.User;
 import org.apache.streampark.console.system.mapper.UserMapper;
 import org.apache.streampark.console.system.service.MemberService;
 import org.apache.streampark.console.system.service.MenuService;
+import org.apache.streampark.console.system.service.RoleService;
 import org.apache.streampark.console.system.service.UserService;
 
 import com.baomidou.mybatisplus.core.conditions.query.LambdaQueryWrapper;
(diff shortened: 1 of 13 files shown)


streampark-console/streampark-console-webapp/src/store/modules/user.js
```diff
@@ -61,6 +61,10 @@ const user = {
       storage.set(USER_ROUTER, routers)
       state.routers = routers
```
1996fanrui, line 72: Should we use sessionStorage here? Mutil page shouldn't share the same router, right?

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/service/impl/UserServiceImpl.java
```diff
@@ -203,6 +210,39 @@ public List<User> findByAppOwner(Long teamId) {
...
+        user.dataMasking();
+        userInfo.put("user", user);
+
+        // 3) roles
+        Set<String> roles = this.roleService.getUserRoleName(username);
+        userInfo.put("roles", roles);
+
+        // 4) permissions
+        Set<String> permissions = this.getPermissions(username);
+        userInfo.put("permissions", permissions);
```
1996fanrui, line 241: Could we return routers here? And remove getRouter.

As I understand, user, roles, permissions and routers should be return together after team is changed.

wolfboys: > Could we return routers here? And remove getRouter.
> 
> As I understand, user, roles, permissions and routers should be return together after team is changed.

The latest routers need to request another interface, It is not returned with these bindings

streampark-console/streampark-console-webapp/src/store/modules/user.js
wolfboys: > Should we use sessionStorage here? Mutil page shouldn't share the same router, right?

oh, yes. you are right.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/service/impl/UserServiceImpl.java
1996fanrui: Why the routers cannot be return here? could you help share more information?

wolfboys: > Why the routers cannot be return here? could you help share more information?

you can see: `getUserRouters` method in `MenuController` for more information.

wolfboys:
hi @1996fanrui  all done. PTAL again

1996fanrui (approved):
@wolfboys thanks for your contribution.

LGTM, however, we should refactor the front-end and back-end about getRouters in the future.

Current code is wired. The routers should be returned to front-end in the `setTeam` interface together.

GuoNingNing (approved):
LGTM

Issue #1423, opened 2022-08-22
[Bug] Incorrect task status after batch task execution is completed


MonsterChenzhuo:
### Search before asking

- [X] I had searched in the [issues](https://github.com/streamxhub/streamx/issues?q=is%3Aissue+label%3A%22bug%22) and found no similar issues.

### What happened
Description: When I run the demo task (batch job), I find that it runs successfully, but the end status is wrong

<img width="1077" alt="图片" src="https://user-images.githubusercontent.com/60029759/185825396-0c11bdef-33e9-4e46-b59f-3a5bba275a76.png">

### StreamX Version

dev

### Java Version

1.8

### Flink Version

1.14.5

### Scala Version of Flink

2.12

### Error Exception

```log
None
```

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

MonsterChenzhuo:
Implementation logic.
We get the corresponding archive log file from the jobid [the log file is generated by the jobmanage at the end]
The archived log file is parsed by json to get the state of the file

<img width="1235" alt="图片" src="https://user-images.githubusercontent.com/60029759/195782312-09997fa9-a3b0-4678-b7af-d10ce212bb63.png">


{
    "jobs":[
        {
            "jid":"002c63091ecc95bd395260dbbdc69d2a",
            "name":"SQLJob-passport-api_kp_ziroom_com_dpLogin-1594315680",
            "state":"FAILED",
            "start-time":1594315681845,
            "end-time":1594315691522,
            "duration":9677,
            "last-modification":1594315691522,
            "tasks":{
                "total":2,
                "created":0,
                "scheduled":0,
                "deploying":0,
                "running":0,
                "finished":0,
                "canceling":0,
                "canceled":1,
                "failed":1,
                "reconciling":0
            }
        }
    ]
}

```java
private def inferApplicationFlinkJobStateFromK8sEvent(@Nonnull trackId: TrackId)
                                                       (implicit pollEmitTime: Long): Option[JobStatusCV] = {

    // infer from k8s deployment and event
    val latest: JobStatusCV = trackController.jobStatuses.get(trackId)
    logger.info(s"Query the local cache result:${trackController.canceling.has(trackId).toString},trackId ${trackId.toString}.")
    val jobState = {
[...]

PR #1877, opened 2022-10-20
[bug]Resolve the problem that the flink jobid stored in the trackid i…

MonsterChenzhuo:

Issue Number: Involving #1423 

## Brief change log

For the precise determination of the state logic, we are using the archive log of jobmanage, but the query of the archive log has become a problem, because the previous logic, trackid the object, the value of jobid is null, resulting in you can not go to the corresponding query, so propose the current pr

## Verifying this change

### DIFF
--- a/streampark-common/src/main/scala/org/apache/streampark/common/conf/ConfigConst.scala
+++ b/streampark-common/src/main/scala/org/apache/streampark/common/conf/ConfigConst.scala
@@ -45,11 +45,7 @@ object ConfigConst {
   val KEY_PASSWORD = "password"
 
   val KEY_TIMEOUT = "timeout"
-
-  val KEY_JOB_ID = "jobId"
-
-  val KEY_FLINK_JOB_ID = "flinkJobId"
-
+  
   val KEY_SEMANTIC = "semantic"
 
   /**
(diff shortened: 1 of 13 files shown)


MonsterChenzhuo:
hi @wolfboys ，I have made the update and can continue with the code review

wolfboys:
> hi @wolfboys ，I have made the update and can continue with the code review

thanks for your contribution, I'll review it later

streampark-common/src/main/scala/org/apache/streampark/common/conf/ConfigConst.scala
```diff
@@ -48,6 +48,8 @@ object ConfigConst {
 
   val KEY_JOB_ID = "jobId"
 
+  val KEY_FLINK_JOB_ID = "flinkJobId"
```
wolfboys: hi MonsterChenzhuo:
I recommend to add `id` and `jobId` field in `SubmitRequest`, Don't add it to the global constant

MonsterChenzhuo:
@wolfboys Thanks for the huge review work, I have revised all your suggestions.

wolfboys (approved):
LGTM

PR #1881, opened 2022-10-21
[Feature] Check apps and projects before delete team

### LINKED ISSUE #1880: [Feature] Check apps and projects before delete team
### Search before asking

### Description

Check apps and projects before delete team

### Usage Scenario

_No response_

### Related issues

_No response_

### Code of Conduct

1996fanrui:

Issue Number: close #1880 

Check apps and projects before delete team

## Brief change log

Check apps and projects before delete team

## Verifying this change

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/service/impl/TeamServiceImpl.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/service/impl/TeamServiceImpl.java
@@ -19,10 +19,16 @@ package org.apache.streampark.console.system.service.impl;
 
 import org.apache.streampark.common.util.AssertUtils;
 import org.apache.streampark.console.base.domain.RestRequest;
+import org.apache.streampark.console.base.exception.ApiAlertException;
 import org.apache.streampark.console.core.enums.UserType;
+import org.apache.streampark.console.core.service.ApplicationService;
+import org.apache.streampark.console.core.service.CommonService;
+import org.apache.streampark.console.core.service.ProjectService;
+import org.apache.streampark.console.core.service.VariableService;
 import org.apache.streampark.console.system.entity.Team;
 import org.apache.streampark.console.system.entity.User;
 import org.apache.streampark.console.system.mapper.TeamMapper;
+import org.apache.streampark.console.system.service.MemberService;
 import org.apache.streampark.console.system.service.TeamService;
 import org.apache.streampark.console.system.service.UserService;
 
@@ -47,6 +53,21 @@ public class TeamServiceImpl extends ServiceImpl<TeamMapper, Team> implements Te
 
     @Autowired
     private UserService userService;
+
+    @Autowired
+    private ApplicationService applicationService;
+
+    @Autowired
+    private ProjectService projectService;
+
+    @Autowired
+    private MemberService memberService;
+
+    @Autowired
+    private VariableService variableService;
+
+    @Autowired
+    private CommonService commonService;
 
     @Override
     public IPage<Team> findTeams(Team team, RestRequest request) {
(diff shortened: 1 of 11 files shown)


streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/service/impl/TeamServiceImpl.java
```diff
@@ -74,9 +94,21 @@ public void createTeam(Team team) {
...
 
     @Override
-    public void deleteTeam(Team team) {
-        // TODO 查询 app 和 project，如果还有 app 或 project 没有删除，则 team 不能被删除。
-        this.removeById(team);
+    public void deleteTeam(Long teamId) {
+        log.info("{} Proceed delete team[Id={}]", commonService.getCurrentUser().getUsername(), teamId);
+        Team team = this.getById(teamId);
+        if (team == null) {
+            log.info("The team[Id={}] doesn't exists.", teamId);
```
wolfboys: I recommend use `ApiAlertException`

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/service/impl/TeamServiceImpl.java
```diff
@@ -74,9 +94,21 @@ public void createTeam(Team team) {
...
+            return;
+        }
+        AssertUtils.state(applicationService.countByTeamId(teamId) == 0,
+            String.format("Please delete the applications under the team[name=%s] first!", team.getTeamName()));
```
wolfboys: I recommend use `ApiAlertException`

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/service/impl/TeamServiceImpl.java
1996fanrui: Thanks for your review, updated.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/service/impl/TeamServiceImpl.java
```diff
@@ -74,9 +95,23 @@ public void createTeam(Team team) {
...
+        }
+        if (applicationService.countByTeamId(teamId) > 0) {
+            throw new ApiAlertException(String.format("Please delete the applications under the team[name=%s] first!", team.getTeamName()));
+        }
+        if (projectService.countByTeamId(teamId) > 0) {
+            throw new ApiAlertException(String.format("Please delete the projects under the team[name=%s] first!", team.getTeamName()));
+        }
+
+        memberService.deleteByTeamId(teamId);
+        variableService.deleteByTeamId(teamId);
```
wolfboys: Delete the variables created under the team directly without reminding the user. I am worried that this will cause trouble to the user's operation. I suggest that before deleting, if any variable under the team, we can ask the user whether to confirm the deletion, After get the user's confirmation then delete the operation, may be a better way

1996fanrui: This team doesn't have any project and application, so these variables must not be used. Why can't delete them directly?

If StreamPark can't delete variables directly, why the member mapping can be deleted directly?

wolfboys: > This team doesn't have any project and application, so these variables must not be used. Why can't delete them directly?
> 
> If StreamPark can't delete variables directly, why the member mapping can be deleted directly?

Variables have a separate management module. Variables essentially belong to "resources". Resources include project, application, etc., as well as the management of jar and udf that we will do later. I think the deletion of all resources must be confirmed by the user at least, and member Just made a relationship mapping, I think it is reasonable to delete member directly

1996fanrui: Thanks for your clarification. updated.

wolfboys (approved):
LGTM

PR #1927, opened 2022-10-30
[Fix] fix application create bug and add variable preview

wangsizhu0504:

fix application create and update bug 

add variable preview

iShot_2022-10-30_22 35 09


wolfboys (approved):
LGTM. good job! Thanks for your hard work

PR #1934, opened 2022-10-31
[Improve]: change alert style and fix suggestion bug

wangsizhu0504:

1. Solve the problem that the suggestion list will have multiple duplicate values when repeatedly entering the application screen
2. Alert validation when creating an application
3. Updated variable interface


wolfboys (approved):
LGTM

PR #1957, opened 2022-11-03
[Fix] handle history version error,improve login footer style

wangsizhu0504:

1.change login footer style 
2. fix update application, select history version bug


wolfboys (approved):
LGTM

PR #1974, opened 2022-11-06
[Improve]: add npm registry and change router menu logic

wangsizhu0504:

1. Add a local npm repository and download the dependencies from the npm mirror source by default when installing
```bash

# .npmrc
registry = 'https://registry.npmmirror.com'
```

2.Unified component file name specification, default Pascal nomenclature
iShot_2022-11-07_00 16 12

3.Adjust the menu logic so that the parent menu will not show the drop-down box when the default child route is set to hidden

iShot_2022-11-07_00 18 05

4.Remove menu creation, update function, maintained by backend staff

> When there are multiple click permissions under the menu, you need to add a click permission of view to control whether you have the permission to view the page

### DIFF
--- a/streampark-console/streampark-console-newui/.gitignore
+++ b/streampark-console/streampark-console-newui/.gitignore
@@ -1,7 +1,6 @@
 node_modules
 .DS_Store
 dist
-.npmrc
 .cache
 
 tests/server/static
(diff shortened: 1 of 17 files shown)


wolfboys (approved):
LGTM

PR #1980, opened 2022-11-07
[Fix] updating podTemplate bug and  change start log api params

wangsizhu0504:
1.Repair the content is empty when updating podTemplate
2.  change start log api params


MonsterChenzhuo:
LGTM

MonsterChenzhuo (approved):

PR #1986, opened 2022-11-08
[bug]batch job end ingress is not deleted

MonsterChenzhuo:

Issue Number: close #xxx 

## Verifying this change

*(or)*

*(or)*


streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/K8sFlinkChangeEventListener.java
```diff
@@ -165,6 +165,7 @@ private void updateApplicationWithJobStatusCV(Application app, JobStatusCV jobSt
         app.setJobId(jobStatus.jobId());
         app.setTotalTask(jobStatus.taskTotal());
         if (FlinkJobState.isEndState(state)) {
+            IngressController.deleteIngress(app.getJobName(), app.getK8sNamespace());
```
wolfboys: I have two questions:
1) The judgment here can be combined with line 182.
2) I don't know if FlinkJobState.isEndState == true, is need to call `IngressController.deleteIngress`? if right, You can check it.

wolfboys (approved):
LGTM

PR #1987, opened 2022-11-08
[Fix] Modify role menu initialization sql,Role menu verification

wangsizhu0504:
1.Remove menu create update menu permission
2.Modify role menu initialization sql

	The view permission is selected by default in the initialization menu,

iShot_2022-11-08_16 14 19


wolfboys (approved):
LGTM

PR #1988, opened 2022-11-08
[Fix] wrong action type in k8s submit action log

rinkako:
There some log with wrong action hint in K8S submission causes puzzle, action `submit` is wrongly writen as `stop`.


wolfboys (approved):
LGTM, Thanks for your contribution, we are looking forward to your next contribution

PR #1996, opened 2022-11-09
[Improve] Add i18n support for system user menu.

green241:
Add i18n support for system user menu.


wangsizhu0504:
Thank you for your contribution. We have modified the i18 of the user and menu section. If you are interested, you can try to modify the i18n of the application module. We apologize for the inconvenience caused!

green241:
Oka, I see.

PR #2005, opened 2022-11-11
[Fix] username len and fix role error

wangsizhu0504:

Issue Number: close #xxx 

## Verifying this change

*(or)*

*(or)*

Modified the user name length limit error issue when creating a user

add project:delete role and app:lauch role


streampark-console/streampark-console-webapp/src/locales/lang/zh-CN/system/user.ts
```diff
@@ -39,7 +39,7 @@ export default {
     lastLoginTime: '最近登录时间',
     password: '密码',
     passwordRequire: '密码为必填项',
-    passwordHelp: '密码长度不能少于 8 个字符',
+    passwordHelp: '密码长度不能少于 20 个字符',
```
wolfboys: "不能大于20个字符"

wangsizhu0504: Sorry, thanks for the reminder. There is a mistake here. The password length cannot be less than 8 characters

1996fanrui (commented):
Hi @wangsizhu0504 , why do you add the `pnpm-lock.yaml`?

wangsizhu0504:
> Hi @wangsizhu0504 , why do you add the `pnpm-lock.yaml`? 

When there is a lock file, the front-end installation dependency will be downloaded according to the information recorded in the lock file, which records the version of the dependency installation, which can sometimes avoid some unnecessary problems. At the same time, you can also speed up your dependency on downloads. If there is no lock.yam, npm will download the latest version of the dependency, but sometimes the latest version of the dependency may affect the project.

wolfboys:
> > Hi @wangsizhu0504 , why do you add the `pnpm-lock.yaml`?
> 
> When there is a lock file, the front-end installation dependency will be downloaded according to the information recorded in the lock file, which records the version of the dependency installation, which can sometimes avoid some unnecessary problems. At the same time, you can also speed up your dependency on downloads. If there is no lock.yam, npm will download the latest version of the dependency, but sometimes the latest version of the dependency may affect the project.

We can refer other apache projects how to solve this problem, such as Apache Flink

streampark-console/streampark-console-webapp/README.md
```diff
@@ -49,6 +49,19 @@ Open the project in Gitpod (free online dev environment for GitHub) and start co
 
 Console is written using Javascript. If you don't have a Node.js development environment, please [set it up](https://nodejs.org/en/download/). The minimum version required is 14.0.
 
+- Set the npm mirror source If you are experiencing a slow download of dependencies, you need to configure the npm mirror source by finding .npmrc in the user folder and writing the following
+
+```bash
+strict-peer-dependencies=false
+save-workspace-protocol = rolling
+registry = 'http://registry.npm.taobao.org'
+sass_binary_site='http://npm.taobao.org/mirrors/node-sass/'
```
wolfboys: hi sizhu:
`http://registry.npm.taobao.org` Chinese developers access npm faster , so we can declare in README.zh-CN.md not README.md(developers from other countries don't know it. I recommend to set it as the official mirror of npm)

wangsizhu0504:
> > > Hi @wangsizhu0504 , why do you add the `pnpm-lock.yaml`?
> > 
> > 
> > When there is a lock file, the front-end installation dependency will be downloaded according to the information recorded in the lock file, which records the version of the dependency installation, which can sometimes avoid some unnecessary problems. At the same time, you can also speed up your dependency on downloads. If there is no lock.yam, npm will download the latest version of the dependency, but sometimes the latest version of the dependency may affect the project.
> 
> We can refer other apache projects how to solve this problem, such as Apache Flink

ok! Well, it was thoughtless before. The current solution is to clear the lock file (the official source of NPM in China is slow to download), and add the method of setting the image source in readme. Chinese developers can speed up dependent downloads according to this method when using this project

streampark-console/streampark-console-webapp/README.md
wangsizhu0504: ok！Thanks for reminding

wolfboys (approved):
LGTM

PR #2006, opened 2022-11-11
[Improve]Add i18n support for flink app menu (add/edit).

green241:
[Improve]Add i18n support for flink app menu (add/edit).


wolfboys:
hi green241:
thanks for your hard work, we will review it leater.

wolfboys (approved):
LGTM

PR #2010, opened 2022-11-12
[improve] Table maximized, modifying table style

wangsizhu0504:
The table full screen changes to maximized, modifying the table style

iShot_2022-11-12_17 48 32


wolfboys (approved):
LGTM

PR #2011, opened 2022-11-12
[fix] Remove hive sql statement validate.

ziqiang-wang:
## Remove hive sql statement validate.

Issue Number: no

## Brief change log

The syntax verification of hive sql is temporarily removed and can be repaired only when a solution is available.

This is only a temporary solution. If we do not have a better solution, I suggest incorporating this PR. Otherwise, users will not be able to use hive sql statements because some hive sql statements will not pass verification.

## Verifying this change

I have verifying this change in my local environment.


wolfboys:
thanks for your contribution, I will review it later

wolfboys (approved):
LGTM

PR #2046, opened 2022-11-17
[Bug] alarm test bug fixed

wolfboys:
[Bug] alarm test bug fixed

Issue Number: close #xxx 

## Verifying this change

*(or)*

*(or)*

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/enums/AlertType.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/enums/AlertType.java
@@ -17,14 +17,6 @@
 
 package org.apache.streampark.console.core.enums;
 
-import org.apache.streampark.console.core.service.alert.AlertNotifyService;
-import org.apache.streampark.console.core.service.alert.impl.DingTalkAlertNotifyServiceImpl;
-import org.apache.streampark.console.core.service.alert.impl.EmailAlertNotifyServiceImpl;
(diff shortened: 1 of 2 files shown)


1996fanrui:
Could you add som background? I don't know why need to change these code.

wolfboys:
> Could you add som background? I don't know why need to change these code.

you can see: `AlertServiceImpl` line 81 for more detail.

1996fanrui:
> > Could you add som background? I don't know why need to change these code.
> 
> you can see: `AlertServiceImpl` line 81 for more detail.

After I see, I think current code can fix bug. However, I don't think it's a good solution.

The code is very unclear. Except for the developer himself, no one else may know that the name is used to find the service. That why you write a bug in #2023 

Clear code can reduce a lot of bugs, don't write a lot of implicit constraints.

1996fanrui:
I've seen too many bugs caused by implicit constraints in StreamPark. Every logic should have some code to implement, instead of implementing features by naming.

Long-term tolerance of low-quality code will make the project very difficult to maintain in the future, and very prone to bugs.

wolfboys:
> I've seen too many bugs caused by implicit constraints in StreamPark. Every logic should have some code to implement, instead of implementing features by naming.
> 
> Long-term tolerance of low-quality code will make the project very difficult to maintain in the future, and very prone to bugs.

I agree with your point of view. The work of code cleaning will be carried out after the release of the first apache version, and later as more and more contributors, the quality will be strictly required.

wolfboys:
cc @1996fanrui PTAL again, thanks

1996fanrui (approved):
Thanks for your update.

LGTM

PR #2056, opened 2022-11-18
[bug] when execute mysql-data.sql file first, it appears an error 'Error Code: 1064. You have an error in your SQL syntax' . 

green241:
PS: 
Mysql-version:8.0.28 MySQL Community Server.
Execute tools: workbench and client.

image


wolfboys (commented):
It is not recommended to delete. you can be changed to the correct comment format, 
pg and h2 also need to be changed

green241:
> It is not recommended to delete. you can be changed to the correct comment format, pg and h2 also need to be changed

oka.

PR #2063, opened 2022-11-19
[improve] job submit parameter improvement

wolfboys:
[improve] job submit parameter improvement

Currently, when submitting a flink job in streampark, there are three properties, namely `option`, `properties`, and `extraParameter`, which can cause confusion and unclear usage during method invocation, so we have simplified this part by removing the option property and putting all flink The standard configuration of flink is put in `properties`, and the extra parameters are put in `extraParameter`

Issue Number: close #xxx 

## Verifying this change

*(or)*

*(or)*

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/AppBuildPipeServiceImpl.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/AppBuildPipeServiceImpl.java
@@ -445,16 +445,13 @@ public class AppBuildPipeServiceImpl
             return Maps.newHashMap();
         }
         LambdaQueryWrapper<AppBuildPipeline> queryWrapper = new LambdaQueryWrapper<AppBuildPipeline>()
-            .select(AppBuildPipeline::getAppId, AppBuildPipeline::getPipeStatusCode)
             .in(AppBuildPipeline::getAppId, appIds);
-        List<Map<String, Object>> rMaps = baseMapper.selectMaps(queryWrapper);
+
+        List<AppBuildPipeline> rMaps = baseMapper.selectList(queryWrapper);
         if (CollectionUtils.isEmpty(rMaps)) {
             return Maps.newHashMap();
         }
-
-        return rMaps.stream().collect(Collectors.toMap(
-            e -> (Long) e.get("app_id"),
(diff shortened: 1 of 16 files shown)


wolfboys:
cc @1996fanrui @MonsterChenzhuo PTAL, thanks

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/AppBuildPipeServiceImpl.java
```diff
@@ -445,16 +445,13 @@ public Map<Long, PipelineStatus> listPipelineStatus(List<Long> appIds) {
             return Maps.newHashMap();
         }
         LambdaQueryWrapper<AppBuildPipeline> queryWrapper = new LambdaQueryWrapper<AppBuildPipeline>()
-            .select(AppBuildPipeline::getAppId, AppBuildPipeline::getPipeStatusCode)
             .in(AppBuildPipeline::getAppId, appIds);
-        List<Map<String, Object>> rMaps = baseMapper.selectMaps(queryWrapper);
+
+        List<AppBuildPipeline> rMaps = baseMapper.selectList(queryWrapper);
```
muchunjin: Should the name of the variable `rMaps` be changed?

muchunjin (commented):
@wolfboys Except that the naming of the variable `rMaps` needs to be standardized, I have not found any other problems. I will test it as a whole after the merger

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/AppBuildPipeServiceImpl.java
wolfboys: > @wolfboys Except that the naming of the variable `rMaps` needs to be standardized, I have not found any other problems. I will test it as a whole after the merger

thanks for your review. I will improve what you mentioned

wolfboys:
> @wolfboys Except that the naming of the variable `rMaps` needs to be standardized, I have not found any other problems. I will test it as a whole after the merger

done. thanks for your review

MonsterChenzhuo:
LGTM

MonsterChenzhuo (approved):

PR #2065, opened 2022-11-19
[Fix] Alert setting api error

wangsizhu0504:

fix alert setting api use error

iShot_2022-11-19_13 45 03


wolfboys (approved):
LGTM

PR #2086, opened 2022-11-23
[Improve]Add i18n support for flink setting alert tab.

green241:
[Improve]Add i18n support for flink setting alert tab.


wolfboys (approved):
LGTM

PR #2106, opened 2022-11-28
Fix the problem that mybatis plus sets jdbcTypeForNull to null

avatarTaier:
mybatisConfiguration.setJdbcTypeForNull(null);  Modify to mybatisConfiguration. setJdbcTypeForNull (JdbcType. NULL);


avatarTaier:
mybatisConfiguration.setJdbcTypeForNull(null); Modify to mybatisConfiguration. setJdbcTypeForNull (JdbcType. NULL);

wolfboys (approved):
LGTM

PR #2119, opened 2022-12-01
[Improve] cluster id tips i18n support.

green241:
[Improve] In k8s execute mode ,cluster id tips i18n support.


wolfboys (approved):
LGTM

PR #2154, opened 2022-12-13
[polish] fix resource leak; avoid NullPointException ; fix Long equals to int statement ;

GOODBOY008:
[polish] fix resource leak; avoid nullPointException ; fix Long equals to int statement ;


GOODBOY008:
@wolfboys Please code review . If you are availiable.

wolfboys (approved):
LGTM

PR #2175, opened 2022-12-16
[Improve] webui async function  improve

wolfboys:
[Improve] webui async function  improve

### DIFF
--- a/streampark-console/streampark-console-webapp/src/views/flink/app/Add.vue
+++ b/streampark-console/streampark-console-webapp/src/views/flink/app/Add.vue
@@ -121,7 +121,7 @@
       }
     } else {
       openConfDrawer(false);
-      await setFieldsValue({ isSetConfig: false, configOverride: null });
+      setFieldsValue({ isSetConfig: false, configOverride: null });
     }
   }
 
@@ -225,7 +225,7 @@
     }
 
     let config = values.configOverride;
-    if (config != null && config.trim() != '') {
+    if (config != null && config !== undefined && config.trim() != '') {
       config = encryptByBase64(config);
     } else {
       config = null;
@@ -275,7 +275,7 @@
     const param = {};
     for (const k in params) {
       const v = params[k];
-      if (v != null) {
+      if (v != null && v !== undefined) {
         param[k] = v;
       }
     }
--- a/streampark-console/streampark-console-webapp/src/views/flink/app/Add.vue
+++ b/streampark-console/streampark-console-webapp/src/views/flink/app/Add.vue
@@ -292,7 +292,7 @@
   }
 
   onMounted(async () => {
-    await handleInitForm();
+    handleInitForm();
   });
 </script>
 
--- a/streampark-console/streampark-console-webapp/src/views/flink/app/Add.vue
+++ b/streampark-console/streampark-console-webapp/src/views/flink/app/Add.vue
@@ -136,7 +136,7 @@
         jar: path,
       });
       uploadLoading.value = false;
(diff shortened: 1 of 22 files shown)


wolfboys:
cc @wangsizhu0504 PTAL, thanks

streampark-console/streampark-console-webapp/src/views/flink/app/Add.vue
```diff
@@ -122,7 +121,7 @@
       }
     } else {
       openConfDrawer(false);
-      setFieldsValue({ isSetConfig: false, configOverride: null });
+      await setFieldsValue({ isSetConfig: false, configOverride: null });
```
wangsizhu0504: It seems that there is no need to add await here. First of all, setFieldValue is not an asynchronous function and the function is to assign a value and does not need to be carried out asynchronously. Secondly, there is no code logic behind setFieldValue, and there is no need to await here.

streampark-console/streampark-console-webapp/src/views/flink/variable/components/VariableDrawer.vue
```diff
@@ -150,7 +150,7 @@
     async (data: Recordable) => {
       try {
         variableId.value = null;
-        resetFields();
+        await resetFields();
```
wangsizhu0504: ResetFieldValue is not an asynchronous function. It is not recommended to add the await keyword. In fact, it is just an assignment function and does not need to be performed asynchronously

streampark-console/streampark-console-webapp/src/views/flink/app/EditStreamPark.vue
```diff
@@ -305,7 +304,7 @@
       project: app.projectId,
       ...defaultFormValue,
     });
-    nextTick(() => {
+    await nextTick(() => {
```
wangsizhu0504: This is the content of editor settings. It should not affect the following logic, and it is not recommended to add await

wangsizhu0504 (dismissed):
LGTM

streampark-console/streampark-console-webapp/src/views/flink/app/Add.vue
```diff
@@ -227,7 +225,7 @@
     }
 
     let config = values.configOverride;
-    if (config != null && config !== undefined && config.trim() != '') {
+    if (config != null && config.trim() != '') {
```
wangsizhu0504, line 228: if (config != null && config !== undefined && config.trim() != '') {

streampark-console/streampark-console-webapp/src/views/flink/app/Add.vue
```diff
@@ -277,7 +275,7 @@
     const param = {};
     for (const k in params) {
       const v = params[k];
-      if (v != null && v !== undefined) {
+      if (v != null) {
```
wangsizhu0504, line 278: if (v != null && v !== undefined) {

streampark-console/streampark-console-webapp/src/views/flink/app/EditStreamPark.vue
```diff
@@ -357,7 +356,7 @@
       <template #args="{ model }">
         <ProgramArgs
           ref="programArgRef"
-          v-if="model.args != null && model.args != undefined"
+          v-if="model.args != null"
```
wangsizhu0504, line 359: v-if="model.args != null && model.args != undefined"

wangsizhu0504 (approved):
LGTM

PR #2195, opened 2022-12-28
[Improve] remove jvm-profiler plugin and flame-graph module

wolfboys:
[Improve] remove jvm-profiler plugin and flame-graph module


muchunjin (approved):
LGTM

PR #2208, opened 2022-12-29
[Improve] Add some config to flink-application.conf

ziqiang-wang:
## Brief change log

Add `table.mode` and `execution.checkpointing.tolerable-failed-checkpoints` config to flink-application.conf file.

`table.mode`, It is convenient to configure the execution mode of the task, so that users can avoid the trouble of searching the execution mode parameter key.

`execution.checkpointing.tolerable-failed-checkpoints` default value is 0, In other words, if a checkpoint fails, the mission is restarted. On production, a large value is set to checkpoint failure to prevent task restart.

## Verifying this change

It has been verified in the production environment


wolfboys (approved):
LGTM

PR #2229, opened 2023-01-05
[Bug] fix after flink 1.15 K8S session start&stop error

monrg:

Issue Number:  #2188

1. because after flink 1.15 FlinkKubeClient removed the getRestService method, Therefore, the logic to verify the existence of the getRestService method is added

2.When the streampark shutdown the k8s session network exception, the response returned is null, the actual k8s session is closed, but the streampark is still in the started state, this scenario should be lost state

## Brief change log

fix after flink 1.15 K8S session start&Stop error

## Verifying this change

- *Manually verified the change by testing locally.* -->

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/FlinkClusterServiceImpl.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/FlinkClusterServiceImpl.java
@@ -263,9 +263,7 @@ public class FlinkClusterServiceImpl extends ServiceImpl<FlinkClusterMapper, Fli
         || ExecutionMode.REMOTE.equals(executionModeEnum)) {
       if (ClusterState.STARTED.equals(ClusterState.of(flinkCluster.getClusterState()))) {
         if (!flinkCluster.verifyClusterConnection()) {
-          updateWrapper.set(FlinkCluster::getAddress, null);
-          updateWrapper.set(FlinkCluster::getClusterState, ClusterState.LOST.getValue());
-          update(updateWrapper);
+          updateClusterIsLost(updateWrapper);
           throw new ApiAlertException("current cluster is not active, please check");
         }
       } else {
@@ -299,6 +297,7 @@ public class FlinkClusterServiceImpl extends ServiceImpl<FlinkClusterMapper, Fli
         updateWrapper.set(FlinkCluster::getClusterState, ClusterState.STOPED.getValue());
         update(updateWrapper);
       } else {
+        updateClusterIsLost(updateWrapper);
         throw new ApiAlertException("get shutdown response failed");
       }
     } catch (Exception e) {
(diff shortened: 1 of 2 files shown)


streampark-flink/streampark-flink-submit/streampark-flink-submit-core/src/main/scala/org/apache/streampark/flink/submit/impl/KubernetesNativeSessionSubmit.scala
```diff
@@ -152,8 +152,9 @@ object KubernetesNativeSessionSubmit extends KubernetesNativeSubmitTrait with Lo
       val kubernetesClusterDescriptor = getK8sClusterDescriptorAndSpecification(flinkConfig)
       clusterDescriptor = kubernetesClusterDescriptor._1
       kubeClient = FlinkKubeClientFactory.getInstance.fromConfiguration(flinkConfig, "client")
-
-      if (deployRequest.clusterId != null && kubeClient.getRestService(deployRequest.clusterId).isPresent) {
+      // after flink 1.15 FlinkKubeClient removed the getRestService method
+      if (deployRequest.clusterId != null && hasMethod(kubeClient.getClass, "getRestService")
```
wolfboys, line 156: This code logic is strange,

e.g: `if ( a.hasMethod("bbb") && a.bbb() )`
if a has no bbb method, Next, call the bbb method,  The code compiles will fail.

monrg: The compilation used is 1.14, the actual run will load the corresponding version of the dependency, after 1.15 removed the method getRestService, so it will fail. I will look at the detailed logic of flink again and implement it under optimization

1996fanrui: Thanks @monrg 's contribution.

It's similar to `FlinkClientTrait`, we can add the `FlinkKubeClientTrait` in the `streampark-flink-shims-base` module, it depends flink-1.14 by default.

Every flink-shims extends the `FlinkKubeClientTrait` and call the `FlinkKubeClient#getRestService before flink 1.15` or the `FlinkKubeClient#getService since flink-1.15`.

And KubernetesNativeSessionSubmit always calls the `FlinkKubeClientTrait#getService`, don't care call `getService` or `getRestService`.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/FlinkClusterServiceImpl.java
```diff
@@ -263,9 +263,7 @@ public void shutdown(FlinkCluster cluster) {
         || ExecutionMode.REMOTE.equals(executionModeEnum)) {
       if (ClusterState.STARTED.equals(ClusterState.of(flinkCluster.getClusterState()))) {
         if (!flinkCluster.verifyClusterConnection()) {
-          updateWrapper.set(FlinkCluster::getAddress, null);
-          updateWrapper.set(FlinkCluster::getClusterState, ClusterState.LOST.getValue());
-          update(updateWrapper);
+          updateClusterIsLost(updateWrapper);
```
1996fanrui, line 266: There are some conflicts here, please rebase the dev branch first, thanks~

PR #2300, opened 2023-02-05
Update version to 2.1.0

wolfboys:
Update version to 2.1.0


wolfboys:
cc @1996fanrui PTAL, thx

muchunjin (approved):
LGTM

1996fanrui (commented):
Thanks for your contribution.

`2.1.0-SNAPSHOT` maybe better? WDYT?

wolfboys:
> Thanks for your contribution.
> 
> `2.1.0-SNAPSHOT` maybe better? WDYT?

I've used `-SNAPSHOT` earlier, I did not see `-SNAPSHOT`  better point, can you tell me why it's better?

wolfboys:
> Thanks for your contribution.
> 
> `2.1.0-SNAPSHOT` maybe better? WDYT?

The current branch is `dev`, which is a branch specially used for development, so I think there is no need to add `-SNAPSHOT`

PR #2327, opened 2023-02-13
Remove AssertUtils to fix license issues

wolfboys:
Remove AssertUtils to fix license issues


wolfboys:
cc @1996fanrui PTAL, thx

streampark-common/src/main/scala/org/apache/streampark/common/util/Utils.scala
```diff
@@ -44,14 +53,20 @@ object Utils {
...
+  def required(expression: Boolean): Unit = {
+    if (!expression) {
+      throw new IllegalArgumentException
+    }
+  }
 
-  def require(requirement: Boolean, message: String): Unit = {
-    if (!requirement) {
-      throw new IllegalArgumentException(s"requirement failed: $message")
+  def required(expression: Boolean, errorMessage: Any): Unit = {
```
1996fanrui, line 62: How about `checkArgument` due to it throw `IllegalArgumentException` when `expression == false`?

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/service/impl/UserServiceImpl.java
```diff
@@ -238,7 +238,6 @@ public List<User> findByAppOwner(Long teamId) {
    */
   @Override
   public Map<String, Object> generateFrontendUserInfo(User user, Long teamId, JWTToken token) {
-    AssertUtils.checkNotNull(user);
```
1996fanrui, line 241: Why remove it instead of add a new method here?

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/service/impl/RoleServiceImpl.java
```diff
@@ -86,7 +86,7 @@ public void deleteRole(Long roleId) {
                 () ->
                     new IllegalArgumentException(String.format("Role id [%s] not found", roleId)));
     List<Long> userIdsByRoleId = memberService.findUserIdsByRoleId(roleId);
-    AssertUtils.isTrue(
```
1996fanrui, line 89: `Utils.required` will throw `IllegalArgumentException`, but old check isn't.

Too many checks are replaced to the `Utils.required` due to release, so it's better to add a TODO here. In the long run, it is unreasonable for all checks to throw `IllegalArgumentException`.

1996fanrui (commented):
Hi @wolfboys , thanks for your contribution. I have left some comments, please take a look in your free time~

streampark-common/src/main/scala/org/apache/streampark/common/util/Utils.scala
wolfboys: > How about `checkArgument` due to it throw `IllegalArgumentException` when `expression == false`?

about the method name, I prefer `required`, shorter and more concise

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/service/impl/RoleServiceImpl.java
wolfboys: > `Utils.required` will throw `IllegalArgumentException`, but old check isn't.
> 
> Too many checks are replaced to the `Utils.required` due to release, so it's better to add a TODO here. In the long run, it is unreasonable for all checks to throw `IllegalArgumentException`.

AssertUtils.isTrue is also throw `IllegalArgumentException`, about the exception, it can be improved in the future

1996fanrui (dismissed):
LGTM

wolfboys:
cc @kezhenxu94 PTAL, if you are free, thanks very much

kezhenxu94 (approved):
Thanks for fixing this! +1

wolfboys:
> Thanks for fixing this! +1

Thanks very much for your review. ✌️

muchunjin (approved):
LGTM

lvshaokang (approved):
LGTM

Issue #2215, opened 2022-12-30
[Feature] Render and display the external links as badges in detail page.


saLeox:
### Search before asking

- [X] I had searched in the [feature](https://github.com/apache/incubator-streampark/issues?q=is%3Aissue+label%3A%22Feature%22) and found no similar feature requirement.

### Description

Render the placeholder using metadata from Flink job, such as {job_id}, {yarn_id}, or {job_name}, and display the badges inside detail page.

### Usage Scenario

When user intend to embed some external link related to Flink.

### Related issues

#2191 

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #2331, opened 2023-02-15
[Feature][Issue-2191][Issue-2215] Support external link

### LINKED ISSUE #2191: [Feature] Maintain external link under Setting module
### Search before asking

### Description

Maintain external link under Setting module, including ADD, UPDATE, LIST, DELETE operation.

image

### Usage Scenario

When user intend to embed some external link related to Flink.

### Related issues

_No response_

### Code of Conduct

saLeox:

Issue Number: close #2191  #2215 

## Brief change log
There should be a place to hold and display the useful external link (such as Job Metrics / Savepoint root path, etc) for each Flink Job.

This feature will allow user to customise their own external link, display it as a badge inside the Flink Application detail page, and jump to other page after clicking it.

For example, I am using Grafana, the link url format will be fixed in the following way:
[https://grafana/{template_name}/flink-monitoring?var-JobId=var-JobId={job_id}](https://grafana/%7Btemplate_name%7D/flink-monitoring?var-JobId=var-JobId=%7Bjob_id%7D)

The value of place holder, job_id, can be fetched from existing meta data of Flink Job easily.

Outcome:
image

image

image

The links are bound with each Flink Job, and the place holders are replaced automatically with actual value.

## Verifying this change


wolfboys:
Thanks for your contribution, it looks good, I'll review it later

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/ExternalLinkServiceImpl.java
```diff
@@ -0,0 +1,124 @@
...
+
+  @Override
+  public void delete(Long linkId) {
+    baseMapper.deleteById(linkId);
+  }
+
+  @Override
+  public List<ExternalLink> render(Long appId) {
+    Application app = applicationService.getById(appId);
+    Utils.required(app != null, "Application doesn't exist");
```
wolfboys: Utils.notNull(app, msg...)

streampark-console/streampark-console-service/src/main/assembly/script/schema/mysql-schema.sql
```diff
@@ -492,5 +492,17 @@ create table `t_alert_config` (
...
 
+-- ----------------------------
+-- Table of t_external_link
+-- ----------------------------
+drop table if exists `t_external_link`;
+CREATE TABLE `t_external_link` (
+  `id` bigint not null auto_increment primary key,
+  `name` varchar(255) collate utf8mb4_general_ci default null,
+  `link_url` varchar(255) collate utf8mb4_general_ci default null,
+  `image_url` varchar(255) collate utf8mb4_general_ci default null,
```
wolfboys: If the image address is base64, the varchar(255) length may be too short

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/entity/ExternalLink.java
```diff
@@ -0,0 +1,52 @@
...
+  private Long id;
+
+  @NotBlank(message = "{required}")
+  private String name;
+
+  @NotBlank(message = "{required}")
+  private String linkUrl;
+
+  @NotBlank(message = "{required}")
+  private String imageUrl;
```
wolfboys: about the badge, it is too complicated for the user to provide the url address. I think a good way is to implement the style of the badge on the front end, and the user only needs to provide the text to be displayed. what do you think?

saLeox: Better to let user maintain the badges on their own,  and we don't need to help persist the any info related to badges. They can generate badge from third party easily, such as https://shields.io/.

wolfboys: > Better to let user maintain the badges on their own, and we don't need to help persist the any info related to badges. They can generate badge from third party easily, such as https://shields.io/.

generate badge from third party I think this still has learning costs. We can build several styles and colors for users to choose, and also support user-defined badge colors. The text in the badge must be provided by the user.

wolfboys: > Better to let user maintain the badges on their own, and we don't need to help persist the any info related to badges. They can generate badge from third party easily, such as https://shields.io/.

When introducing any component and service, as long as it is directly facing to the user use, it is necessary to consider the ease of use and the user's cost of use. The simpler the better

saLeox: Now we enable user to customise the badge in the similar way as  https://shields.io/.
Please help review again.

wolfboys:
please resolved the conflicts, thx

streampark-console/streampark-console-webapp/src/locales/lang/zh-CN/flink/setting.ts
```diff
@@ -179,6 +180,29 @@ export default {
...
+      badgeColorIsRequired: '徽章颜色是必选',
+      linkUrl: '链接',
+      linkUrlPlaceholder: '请输入链接',
+      linkUrlIsRequired: '链接是必填',
+    },
+    operateMessage: {
+      createLinkSuccessful: '成功新建外部链接!',
+      createLinkFailed: '创建外部链接失败!',
+      updateLinkSuccessful: '成功更新外部链接!',
+      updateLinkFailed: '更新外部链接失败!',
```
wolfboys, line 201: I suggest change '外部链接' to '扩展链接'

wolfboys (dismissed):

streampark-console/streampark-console-webapp/src/locales/lang/zh-CN/flink/setting.ts
```diff
@@ -179,6 +180,29 @@ export default {
       clusterId: '集群ID',
     },
   },
+  externalLink: {
+    externalLinkSetting: '外部链接',
```
wolfboys, line 184: how about `扩展链接` ?

asuiah: Sure will update for this.

PR #2340, opened 2023-02-18
[Improve][Issue 2325] Change the corresponding lines about exception throw for fine users Interaction

### LINKED ISSUE #2325: [Bug] No alert msg on FE when updating application with exception.
### Search before asking

### What happened

- I can't sure that the case type was known for community because my glance at the 2.0.0 edition preview.

- No alert msg on FE when updating application with exception (It only affects the user experience and is not strictly a bug.
May be  design missing  or classical bugs in the current edition but not exception interceptor catch mechanism missing )

- A solution to fix the case  
image

After the change:

image

### StreamPark Version

N.A 

### Java Version

1.8

### Flink Version

N.A

### Scala Version of Flink

2.12

### Error Exception

```log
N.A
```

### Screenshots

_No response_

### Code of Conduct

RocMarshal:

Issue Number: close #2325 

## Brief change log
Change the corresponding lines about exception throw for fine users Interaction

## Verifying this change

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/interceptor/UploadFileTypeInterceptor.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/interceptor/UploadFileTypeInterceptor.java
@@ -18,7 +18,6 @@
 package org.apache.streampark.console.base.interceptor;
 
 import org.apache.streampark.common.util.FileUtils;
-import org.apache.streampark.common.util.Utils;
 import org.apache.streampark.console.base.exception.ApiAlertException;
 
 import org.springframework.stereotype.Component;
@@ -43,11 +42,11 @@ public class UploadFileTypeInterceptor implements HandlerInterceptor {
       Map<String, MultipartFile> files = multipartRequest.getFileMap();
       for (String file : files.keySet()) {
         MultipartFile multipartFile = multipartRequest.getFile(file);
-        Utils.notNull(multipartFile);
+        ApiAlertException.throwIfNull(
+            multipartFile, "File to upload can't be null. Upload file failed.");
         boolean fileType = FileUtils.isJarFileType(multipartFile.getInputStream());
-        if (!fileType) {
-          throw new ApiAlertException("illegal file type, Only standard jar files supported");
-        }
+        ApiAlertException.throwIfFalse(
+            fileType, "Illegal file type, Only support standard jar files. Upload file failed.");
       }
     }
     return true;
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/exception/ApiAlertException.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/exception/ApiAlertException.java
@@ -19,6 +19,8 @@ package org.apache.streampark.console.base.exception;
 
 import org.apache.streampark.console.base.domain.ResponseCode;
 
+import java.util.Objects;
+
 /**
  *
  *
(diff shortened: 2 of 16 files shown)


streampark-common/src/main/java/org/apache/streampark/common/util/ExceptionGetter.java
```diff
@@ -0,0 +1,27 @@
...
+package org.apache.streampark.common.util;
+
+import javax.annotation.Nonnull;
+
+/** A interface to define the action to create a exception with specified error message. */
+public interface ExceptionGetter {
+
+  @Nonnull
+  Exception get(String errorMsg);
+}
```
RocMarshal: could we move the `org.apache.streampark.console.base.exception` package into `streampark-common` module ? 
It would be better merge the condition logic and throw exceptions logic , which is useful to unify the behavior of `Utils`. But there's dependency conflict in the packages and modules when I want use throw `ApiAlertException` instead of passing by `ExceptionGetter`.

wolfboys: thanks for your contribution, I don't quite agree with this method, we just need to add method in ApiAlertException and ApiDetailException, such as:

```
  public static void throwIfNull(Object object, String message) {
         if (object == null) {
             throw new ApiAlertException(message);
         }
     }
```
how to use:
```
ApiAlertException.throwIfNull(executionMode, "executionMode cannot be null");
```

Would this be simpler?

RocMarshal: Yes. There's nothing better. 
It's my original strategy, too. 
Thanks for your quick-reply.

wolfboys: > Yes. There's nothing better. It's my original strategy, too. Thanks for your quick-reply.

👌👍

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/interceptor/UploadFileTypeInterceptor.java
```diff
@@ -43,11 +42,10 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
       Map<String, MultipartFile> files = multipartRequest.getFileMap();
       for (String file : files.keySet()) {
         MultipartFile multipartFile = multipartRequest.getFile(file);
-        Utils.notNull(multipartFile);
+        ApiAlertException.throwIfNull(multipartFile, "File to upload can't be null.");
```
wolfboys: If using ApiAlertException, In addition show the err message to user, we also meed to tell user the result of the operation, such as success or failure. 
This message must be simple and precise

wolfboys: > If using ApiAlertException, In addition show the err message to user, we also meed to tell user the result of the operation, such as success or failure. This message must be simple and precise

e.g:         ApiAlertException.throwIfNull(multipartFile, "File to upload can't be null. upload jar failed");

RocMarshal: > 

I'm fine to it.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/mapper/RoleMenuMapper.java
```diff
@@ -20,5 +20,7 @@
 import org.apache.streampark.console.system.entity.RoleMenu;
 
 import com.baomidou.mybatisplus.core.mapper.BaseMapper;
+import org.springframework.stereotype.Component;
 
+@Component
```
wolfboys: Why add this annotation?

RocMarshal: I add it to solve  error-red highlight lines caused by error cache of my IDE.

And I removed it, now.
Thank you.

wolfboys (approved):
LGTM

PR #2346, opened 2023-02-20
[Improve] add i18n support for job sql version compare.

green241:
[Improve] add i18n support for job sql version compare.


wolfboys (approved):
LGTM

PR #2351, opened 2023-02-22
[Improve] add i18n support for variable Management show info page.

green241:
[Improve] add i18n support for variable Management show info page.


wolfboys (approved):
LGTM

PR #2353, opened 2023-02-22
[Fix] Test classes should be under sp package

tisonkun:

1. Move tests under outside packages to `org.apache.streampark`.
2. Remove `RefreshCacheTest` since third party dependency should be self tested.


streampark-console/streampark-console-service/src/test/java/org/apache/streampark/common/util/RegexTest.java
```diff
@@ -28,6 +26,8 @@
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 class RegexTest {
```
tisonkun, line 33: BTW, `RegexTest#classLoader` and `RegexTest#flinkVersion` seem quite tricky and fail locally. How can we verify these tests correctly?

streampark-console/streampark-console-service/src/test/java/org/apache/streampark/console/core/bean/SenderEmailTest.java
```diff
@@ -77,7 +77,7 @@ void alert() {
       mail.setStatus(appState.name());
 
       StringWriter writer = new StringWriter();
-      Map<String, AlertTemplate> out = new HashMap<String, AlertTemplate>();
+      Map<String, AlertTemplate> out = new HashMap<>();
```
tisonkun, line 80: ditto this test fails locally

... we just swallow the exception below.

streampark-console/streampark-console-service/src/test/java/org/apache/streampark/common/util/RegexTest.java
wolfboys: This is really tricky to solve, and the local environment needs to meet the requirements. We can delete this testcase first.

tisonkun: Sounds reasonable. But I don't have time to investigate more. Perhaps open as a follow-up.

streampark-console/streampark-console-service/src/test/java/org/apache/streampark/console/core/bean/SenderEmailTest.java
wolfboys: Because it is necessary to test the send email. It is necessary to manually fill in info such as sending emails, password ... These are private info and cannot be provided

wolfboys: I suggest remove these testcase classes

wolfboys (approved):
LGTM

PR #2354, opened 2023-02-22
[Improve] Skip Docker build for forks

tisonkun:
## Brief change log

This workflow always fails on forks since no DOCKER_USER and DOCKER_TOKEN configured. Actually, it should not run in forks.

BTW, perhaps we should not use GHA to release formal Docker images since it's part of a formal release. See https://github.com/apache/incubator-kvrocks/pull/1247#pullrequestreview-1284549026. But I'd regard it as a differnt topic than what this patch do.


wolfboys (approved):
LGTM

PR #2395, opened 2023-03-03
[Issue-2014][test] Add SpringTestBase class.

### LINKED ISSUE #2014: [Improve] Enhance the tests of 'streampark-console' module.
### Search before asking

### Main Objective in the Improvement：
- Supplement unit test cases of streampark console.

#### Present situation: 
At present, there are few test cases for streampark console based on JUnit5(Jupiter).
To make a long story short for no test cases:
  - As a part of the development specification, test cases are indispensable. 
  - It is not conducive to the long-term evolution of the project. 
  - When changing the current module, it is difficult to find potential problems.

### Choose Test Suites(If the improvement was accepted)

JUnit5 & Mockito [& AssertJ]

#### Why JUNIT5?

1. Released in 2017, JUnit 5 has become more and more popular among Java projects. With quite a lot of new features such as extension model, advanced parameterized test, and better support of lambda functions, JUnit 5 presents more convenience and flexibility to developers.
2. Support integration with many frameworks, such as Spring, Testcontainers, etc
[JUnit5 Official website](https://junit.org/junit5/)

#### About Mock

Why Mockito instead of Powermock?

- Mockito supports Junit5
- Powermock works based on Junit4, https://github.com/powermock/powermock
- Mockito's community is more active than Powermock. https://github.com/mockito/mockito

#### Assertion API
- JUnit5 Assertion API: is the built-in API of JUnit5. If We use 'JUnit5 Assertions' to work, we can reduce some learning costs.
- [AssertJ](https://joel-costigliola.github.io/assertj/index.html): AssertJ can perform the same functions as JUnit5 Assertion, and has many APIs based on stream coding, which makes it more elegant.

### Integration test problems and other module test problems.
Maybe we also need to consider the test suites selection based on the sections.

### E2E test 
There is an [issue](https://github.com/apache/incubator-streampark/issues/1381) described the E2E test plan  

### Any suggestion and idea is appreciated!
- About other test suites, TestNG...
- Or anything important for the improvment.

### Usage Scenario

N.A

### E2E test 
There is an [issue](https://github.com/apache/incubator-streampark/issues/1381) described the E2E test plan  

### Code of Conduct

RocMarshal:

Issue Number: close #2014 

## Brief change log

Add SpringTestBase class.

## Verifying this change

N.A

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/runner/EnvInitializer.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/runner/EnvInitializer.java
@@ -70,8 +70,21 @@ public class EnvInitializer implements ApplicationRunner {
           "^streampark-flink-shims_flink-(1.12|1.13|1.14|1.15|1.16)_(2.11|2.12)-(.*).jar$",
           Pattern.CASE_INSENSITIVE | Pattern.DOTALL);
 
+  // We introduced it for https://github.com/apache/incubator-streampark/issues/2014
+  private boolean shouldSkip() {
+    // Use try ... catch to avoid NPE or AIOB.
+    try {
+      return "test".equals(context.getEnvironment().getActiveProfiles()[0]);
+    } catch (Exception e) {
+      return false;
+    }
+  }
+
   @Override
   public void run(ApplicationArguments args) throws Exception {
+    if (shouldSkip()) {
+      return;
+    }
     String appHome = WebUtils.getAppHome();
     if (appHome == null) {
       throw new ExceptionInInitializerError(
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/runner/EnvInitializer.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/runner/EnvInitializer.java
@@ -45,8 +45,10 @@ import java.io.File;
 import java.io.FileFilter;
 import java.io.IOException;
 import java.nio.file.Files;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Optional;
+import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -61,7 +63,7 @@ public class EnvInitializer implements ApplicationRunner {
 
   @Autowired private SettingService settingService;
 
-  private final Map<StorageType, Boolean> initialized = new ConcurrentHashMap<>(2);
+  private final Set<StorageType> initialized = new HashSet<>(2);
 
   private final FileFilter fileFilter = p -> !".gitkeep".equals(p.getName());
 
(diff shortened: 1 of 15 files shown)


streampark-console/streampark-console-service/src/test/java/org/apache/streampark/base/SpringTestBase.java
```diff
@@ -0,0 +1,126 @@
...
+  @Container
+  public static BrowserWebDriverContainer<?> chrome =
+      new BrowserWebDriverContainer<>()
+          .withCapabilities(new ChromeOptions())
+          .withLogConsumer(new Slf4jLogConsumer(LOG));
+
+  // Use it to mock external services.
+  @Container
+  public static MockServerContainer mockServer =
+      new MockServerContainer(MOCKSERVER_IMAGE).withLogConsumer(new Slf4jLogConsumer(LOG));
```
RocMarshal: In fact, I prefer to move these two suites down to an abstract subclass. If a test class does not use these two suites, but still inherits the current class, it is likely to slow down the startup speed of the test class.

streampark-common/src/main/scala/org/apache/streampark/common/conf/ConfigConst.scala
```diff
@@ -32,6 +32,8 @@ object ConfigConst {
 
   val KEY_APP_HOME = "app.home"
 
+  val KEY_SKIP_INIT_ENV = "env.init.skip"
```
wolfboys: I do not recommend add this parameter, maybe we have better way

streampark-console/streampark-console-service/src/test/java/org/apache/streampark/base/SpringTestBase.java
RocMarshal: image

streampark-console/streampark-console-service/src/test/java/org/apache/streampark/console/core/utils/YarnQueueLabelExpressionTest.java
```diff
@@ -60,7 +62,7 @@ void testOf() {
     assertThat(YarnQueueLabelExpression.of("a").getLabelExpression()).isEmpty();
     assertThat(YarnQueueLabelExpression.of("a").getQueue()).isEqualTo("a");
     assertThatThrownBy(() -> YarnQueueLabelExpression.of("a@"))
-        .isInstanceOf(IllegalArgumentException.class)
```
RocMarshal, line 63: As mentioned offline, I was used to think the test case would be checked in the git CI boot.
However, it didn't do it. Of course, the root-cause is mine previous pr instead of the disabled git ci part. hahahaha~

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/runner/EnvInitializer.java
```diff
@@ -70,8 +70,21 @@ public class EnvInitializer implements ApplicationRunner {
...
+    try {
+      return "test".equals(context.getEnvironment().getActiveProfiles()[0]);
+    } catch (Exception e) {
+      return false;
+    }
+  }
+
   @Override
   public void run(ApplicationArguments args) throws Exception {
+    if (shouldSkip()) {
```
wolfboys: how about?
```
Optional<String> profile = Arrays.stream(context.getEnvironment().getActiveProfiles()).findFirst();
    if ("test".equals(profile.orElse(null))) {
        return;
    }
```

RocMarshal: Thx for the commentt @wolfboys .
I'd like it.
Updated  & made optimization change in the EnvInitializer.
PTAL in your free time~.

wolfboys: > Thx for the commentt @wolfboys . I'd like it. Updated & made optimization change in the EnvInitializer. PTAL in your free time~.

👌 good job.

wolfboys (approved):
LGTM, thanks for your hard work

PR #2403, opened 2023-03-06
[Fix] add permission for /flink/conf/delete

zhoulii:

Streampark adds 'conf:delete' permission in [mysql-data.sql#L83](https://github.com/apache/incubator-streampark/blob/dev/streampark-console/streampark-console-service/src/main/assembly/script/data/mysql-data.sql#L83) , but never use it.

## Verifying this change


wolfboys (approved):
LGTM

PR #2446, opened 2023-03-15
[Fix] remove candidate column in conf tab

zhoulii:

we use effective column to show effective or latest config, see [DetailTab.vue#L350-L353](https://github.com/apache/incubator-streampark/blob/dev/streampark-console/streampark-console-webapp/src/views/flink/app/components/AppDetail/DetailTab.vue#L350-L353), and candidate is not a property of [ApplicationConfig.java](https://github.com/apache/incubator-streampark/blob/dev/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/entity/ApplicationConfig.java), so we can remove candidate column in config tab.

before:
before

after:
after

## Verifying this change


wolfboys (approved):
LGTM

PR #2449, opened 2023-03-16
[fix] move all java code from src/main/scala to src/main/java

zhoulii:

move all java code from `src/main/scala` to `src/main/java`

## Verifying this change

- *Manually verified the change by testing locally.* -->


wolfboys (approved):
LGTM

PR #2452, opened 2023-03-17
[Imporve] fix that savepointing state lasts for a long time

zhoulii:

when we trigger savepoint in streampark web, the app status will change to `SAVEPOINTING`, and `SAVEPOINTING` state lasts for a long time, even if the savepoint operation has already finished.

## Verifying this change


wolfboys (approved):
LGTM

PR #2495, opened 2023-03-23
[improve] support i18n for tab name & add copy button for backup path

zhoulii:

1. add i18n support for tab name

2. add copy button for backup path

3. change `Option Log` to `Operation Log`

## Verifying this change

### DIFF
--- a/streampark-console/streampark-console-webapp/src/locales/lang/zh-CN/flink/app.ts
+++ b/streampark-console/streampark-console-webapp/src/locales/lang/zh-CN/flink/app.ts
@@ -90,7 +90,7 @@ export default {
         option: '选项',
         configuration: '配置',
         flinkSql: 'Flink SQL',
-        savepoint: '保存点',
+        savepoint: 'Savepoint',
         backup: '备份',
         operationLog: '操作日志'
       },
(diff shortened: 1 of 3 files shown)


zhoulii:
cc @wangsizhu0504

streampark-console/streampark-console-webapp/src/locales/lang/zh-CN/flink/app.ts
```diff
@@ -86,6 +86,14 @@ export default {
     nullAccessToken: '访问令牌为空，请联系管理员添加.',
     invalidAccessToken: '访问令牌无效，请联系管理员。',
     detailTab: {
+      detailTabName: {
+        option: '选项',
+        configuration: '配置',
+        flinkSql: 'Flink SQL',
+        savepoint: '保存点',
```
wangsizhu0504: Whether savepoint does not require translation should be a proper term. It is best to discuss this with back-end personnel

wangsizhu0504 (approved):
LGTM!

streampark-console/streampark-console-webapp/src/locales/lang/zh-CN/flink/app.ts
zhoulii: many chinese blogs use ‘保存点’ for savepoint， use '检查点' for checkpoint，but Savepoint is indeed a professional term, maybe use ‘Savepoint’ directly is a better choice.

wolfboys: > Whether savepoint does not require translation should be a proper term. It is best to discuss this with back-end personnel

"savepoint" is not recommended to be translated into Chinese

PR #2508, opened 2023-03-26
[Fix]: build error

wangsizhu0504:

* After updating to vite 4.2, use pnpm to install dependencies, and no errors were reported during compilation. However, the page cannot be accessed. Index.js was not injected in index.html, possibly due to the fact that the imported vite plug-in does not adapt to vite 4.0. Currently, the temporary solution is to rollback to vite 3, and the plug-in will be adapted in the future *


tisonkun:
Can you cross refer to the vite issue here?

wangsizhu0504:
> Can you cross refer to the vite issue here?

https://github.com/vitejs/vite/issues/12232    Problems of the same kind

wolfboys (approved):
LGTM

PR #2516, opened 2023-03-28
[BUG]fix call deleteIngress reverse

Jcodelove:
## Correction of input parameters
This method definition is `def deleteIngress(ingressName: String, nameSpace: String)`
but it is used this way in some places
`IngressController.deleteIngress(application.getK8sNamespace(), application.getJobName());`
so flip the parameter


wolfboys:
cc @MonsterChenzhuo PTAL

MonsterChenzhuo:
LGTM

MonsterChenzhuo (approved):

PR #1585, opened 2022-09-13
[Bug] Fix the wrong duration

1996fanrui:
### What problem does this PR solve?

Issue Number: close #1584 

Problem Summary:

Fix the wrong duration, it should use the `Math.floor`, that is, `Returns the greatest integer less than or equal to its numeric argument.`

image

### What is changed and how it works?

Proposal: [xxx](url) 

Feel free to ping committers for the review!

## Contribution Checklist

  - Make sure that the pull request corresponds to a [GITHUB issue](https://github.com/apache/streampark/issues).

  - Name the pull request in the form "[Feature] [component] Title of the pull request", where *Feature* can be replaced by `Hotfix`, `Bug`, etc.

  - Minor fixes should be named following this pattern: `[hotfix] [docs] Fix typo in README.md doc`.

## Purpose of this pull request

Fix the wrong duration

PR #2523, opened 2023-03-29
[bug] fix the wrong duration value in webui

zhoulii:

when duration time is less than 87ms, function [dateToDuration](https://github.com/apache/incubator-streampark/blob/dev/streampark-console/streampark-console-webapp/src/utils/dateUtil.ts#L38) will return a bad result. This is because `parseInt` cannot handle scientific notation well.

duration_error

**FYI:** same reason as [PR-1585](https://github.com/apache/incubator-streampark/pull/1585)

## Verifying this change


wolfboys:
cc @wangsizhu0504 @1996fanrui PTAL thx

wangsizhu0504 (approved):
LGTM

PR #2524, opened 2023-03-29
[hotfix] Add console module pid write operation back

GOODBOY008:
## Brief change log

Add console module pid write operation back


wolfboys (approved):
LGTM

PR #2531, opened 2023-03-30
[bug] change flinksql to a history version does not take effect

zhoulii:

when editing an application in web-ui, we can change flinksql to a history version, but the selected history version won't take effect. That because the selected `sqlId` passed from frontend is incorrect and would make streampark create a new flinksql in backend whenever the flinksql version is changed.

## Verifying this change


zhoulii:
cc @wangsizhu0504

wangsizhu0504 (approved):
LGTM

Issue #2498, opened 2023-03-23
[Feature] [SubTask] The cluster supports remote and yarn session heartbeat monitoring


xujiangfeng001:
### Search before asking

- [X] I had searched in the [feature](https://github.com/apache/incubator-streampark/issues?q=is%3Aissue+label%3A%22Feature%22) and found no similar feature requirement.

### Description

The cluster supports remote and yarn session heartbeat monitoring and synchronous updates to the database.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

xujiangfeng001:
The following is the specific implementation logic diagram:
![未命名文件 (1)](https://user-images.githubusercontent.com/104614523/227490836-b41d7214-9b33-46cd-a9ef-96bd8c9105ba.png)

PR #2541, opened 2023-03-31
[ISSUE-2498][Feature] [SubTask] The cluster supports remote and yarn session heartbeat monitoring

xujiangfeng001:

Issue Number: https://github.com/apache/incubator-streampark/issues/2498

This pr supports status monitoring and status updates for the "Flink Cluster on Yarn" and "Remote".For the status change of "Flink Cluster" in this pr, see:

未命名文件

Heartbeat and status monitoring support for "Flink Cluster on K8s" will be supported in subsequent pr support.

## Verifying this change

This change  without any test coverage.

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/annotation/CheckApp.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/annotation/CheckApp.java
@@ -1,0 +1,29 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.streampark.console.core.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface CheckApp {
+  String value() default "";
+}
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/annotation/CheckTeam.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/annotation/CheckTeam.java
@@ -1,0 +1,29 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.streampark.console.core.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface CheckTeam {
+  String value() default "";
+}
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/annotation/CheckUser.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/annotation/CheckUser.java
@@ -1,0 +1,29 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.streampark.console.core.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface CheckUser {
+  String value() default "";
+}
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/aspect/StreamParkAspect.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/aspect/StreamParkAspect.java
@@ -22,9 +22,19 @@ package org.apache.streampark.console.core.aspect;
 import org.apache.streampark.console.base.domain.RestResponse;
 import org.apache.streampark.console.base.exception.ApiAlertException;
 import org.apache.streampark.console.core.annotation.ApiAccess;
+import org.apache.streampark.console.core.annotation.CheckApp;
+import org.apache.streampark.console.core.annotation.CheckTeam;
+import org.apache.streampark.console.core.annotation.CheckUser;
+import org.apache.streampark.console.core.entity.Application;
+import org.apache.streampark.console.core.enums.UserType;
+import org.apache.streampark.console.core.service.ApplicationService;
+import org.apache.streampark.console.core.service.CommonService;
 import org.apache.streampark.console.core.task.FlinkRESTAPIWatcher;
 import org.apache.streampark.console.system.entity.AccessToken;
+import org.apache.streampark.console.system.entity.User;
+import org.apache.streampark.console.system.service.MemberService;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.shiro.SecurityUtils;
 
 import lombok.extern.slf4j.Slf4j;
@@ -34,6 +44,11 @@ import org.aspectj.lang.annotation.Aspect;
 import org.aspectj.lang.annotation.Pointcut;
 import org.aspectj.lang.reflect.MethodSignature;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.core.DefaultParameterNameDiscoverer;
+import org.springframework.expression.EvaluationContext;
+import org.springframework.expression.Expression;
+import org.springframework.expression.spel.standard.SpelExpressionParser;
+import org.springframework.expression.spel.support.StandardEvaluationContext;
 import org.springframework.stereotype.Component;
 
 import java.util.Objects;
@@ -44,6 +59,9 @@ import java.util.Objects;
 public class StreamParkAspect {
 
   @Autowired private FlinkRESTAPIWatcher flinkRESTAPIWatcher;
+  @Autowired private CommonService commonService;
+  @Autowired private MemberService memberService;
+  @Autowired private ApplicationService applicationService;
 
   @Pointcut(
       "execution(public"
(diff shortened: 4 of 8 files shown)


RocMarshal:
Nice~! 👍 
@xujiangfeng001 Thank your for the contribution. 
Could you please  help to fill the PR description? I found that some items were set as default.
As you know, the precise and concise description will be helpful to reviewers, which makes sense for advancing the pr. 
Looking forward to it~

xujiangfeng001:
> Nice~! 👍 @xujiangfeng001 Thank your for the contribution. Could you please help to fill the PR description? I found that some items were set as default. As you know, the precise and concise description will be helpful to reviewers, which makes sense for advancing the pr. Looking forward to it~

@RocMarshal Thank you for your suggestion. I have completed the project description.Looking forward to your review again,

RocMarshal:
@xujiangfeng001 Thanks for your update. Nice~
I'll review it asap.

streampark-common/src/main/java/org/apache/streampark/common/enums/ClusterState.java
```diff
@@ -24,12 +24,21 @@ public enum ClusterState implements Serializable {
...
   /** cluster stopped */
   STOPPED(2),
-
   /** cluster lost */
-  LOST(3);
+  LOST(3),
+  /** cluster undefined */
+  UNDEFINED(4),
+  /** cluster successfully */
+  SUCCEEDED(5),
```
RocMarshal: The two enum look like redundant on platform side.
Its could be used to represent the final status on yarn just platform side.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/FlinkClusterServiceImpl.java
```diff
@@ -182,21 +187,25 @@ public void start(FlinkCluster cluster) {
           String address =
               YarnUtils.getRMWebAppURL() + "/proxy/" + deployResponse.clusterId() + "/";
           flinkCluster.setAddress(address);
+          flinkCluster.setJobManagerUrl(deployResponse.address());
```
RocMarshal:
```suggestion
```

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/FlinkClusterServiceImpl.java
```diff
@@ -182,21 +187,25 @@ public void start(FlinkCluster cluster) {
...
         } else {
           flinkCluster.setAddress(deployResponse.address());
         }
         flinkCluster.setClusterId(deployResponse.clusterId());
```
RocMarshal:
```suggestion
        flinkCluster.setJobManagerUrl(deployResponse.address());
        flinkCluster.setClusterId(deployResponse.clusterId());
```

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkClusterWatcher.java
```diff
@@ -0,0 +1,248 @@
...
+import java.util.concurrent.TimeUnit;
+
+/** This implementation is currently used for tracing Cluster on yarn,remote,K8s mode */
+@Slf4j
+@Component
+public class FlinkClusterWatcher {
+
+  @Autowired private FlinkClusterService flinkClusterService;
+
+  private Long lastWatcheringTime = null;
```
RocMarshal: AnchorA: What about the Long.MIN_VALUE on init-value  ?

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkClusterWatcher.java
```diff
@@ -0,0 +1,248 @@
...
+  @PreDestroy
+  private void stop() {
+    // TODO: flinkcluster persistent
+  }
+
+  @Scheduled(fixedDelay = 1000)
+  private void start() {
+    if (lastWatcheringTime == null) {
+      watcher();
+    } else if (System.currentTimeMillis() - lastWatcheringTime >= WATCHER_INTERVAL) {
```
RocMarshal:
```suggestion
   if (System.currentTimeMillis() - lastWatcheringTime >= WATCHER_INTERVAL) {
```
It could be adjusted due to `anchorA`.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkClusterWatcher.java
```diff
@@ -0,0 +1,248 @@
...
+@Slf4j
+@Component
+public class FlinkClusterWatcher {
+
+  @Autowired private FlinkClusterService flinkClusterService;
+
+  private Long lastWatcheringTime = null;
+
+  // Track interval  every 30 seconds
+  private static final Long WATCHER_INTERVAL = 1000L * 30;
```
RocMarshal: What about `  private static final Duration WATCHER_INTERVAL = Duration.ofSeconds(30);` ?

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkClusterWatcher.java
```diff
@@ -0,0 +1,248 @@
...
+          TimeUnit.SECONDS,
+          new LinkedBlockingQueue<>(1024),
+          ThreadUtils.threadFactory("flink-cluster-watching-executor"));
+
+  /** Initialize cluster cache */
+  @PostConstruct
+  private void init() {
+    WATCHER_CLUSTERS.clear();
+    List<FlinkCluster> flinkClusters = flinkClusterService.list();
+    flinkClusters.forEach(cluster -> WATCHER_CLUSTERS.put(cluster.getId(), cluster));
```
RocMarshal: The expected state of some clusters is either stopped or abnormal, which should no longer require monitoring.
What do you think about on it ?



Based on the above questions, imagine a situation where Yarn session1 cluster is in a running state. After running for a period of time, the cluster state is FAILED, while Yarn session2 is a cluster that has been created but not started. However, at this point, the platform restarts. How can we ensure that after the restart, only Yarn session1 continues to be monitored and Yarn session2 is ignored?

For example, introducing an expectState field to identify the expected state of the cluster, only monitoring the heartbeat of clusters with an expected state of RUNNING.

I'd like to hear anything better about solving the  cases.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkClusterWatcher.java
```diff
@@ -0,0 +1,248 @@
...
+      EXECUTOR.execute(
+          () -> {
+            FlinkCluster flinkCluster = entry.getValue();
+            Integer clusterExecutionMode = flinkCluster.getExecutionMode();
+            if (!ExecutionMode.isKubernetesSessionMode(clusterExecutionMode)) {
+              ClusterState state = getClusterState(flinkCluster);
+              handleClusterState(flinkCluster, state);
+            } else {
+              // TODO: K8s Session status monitoring
+            }
```
RocMarshal: What about splitting its into three methods mapping to the corresponding three cases `remote`,`y-session`, `k8s-session` ?

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkClusterWatcher.java
RocMarshal:
```suggestion
  private Long lastWatchingTime = Long.MIN_VALUE;
```

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkClusterWatcher.java
```diff
@@ -0,0 +1,248 @@
...
+    }
+    String yarnUrl = "ws/v1/cluster/apps/".concat(flinkCluster.getClusterId());
+    try {
+      String result = YarnUtils.restRequest(yarnUrl);
+      if (null == result) {
+        return ClusterState.UNKNOWN;
+      }
+      YarnAppInfo yarnAppInfo = JacksonUtils.read(result, YarnAppInfo.class);
+      ClusterState state = ClusterState.of(yarnAppInfo.getApp().getFinalStatus());
+      if (ClusterState.UNDEFINED.equals(state)) {
```
RocMarshal: The Yarn resourcemanager rest API will include special cases where the state is accepted but the final status is UNDEFINED.

So it would be better  to judge the current Yarn session status based on `state` and `finalStatus` of yarnAPP.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkClusterWatcher.java
```diff
@@ -0,0 +1,248 @@
...
+  private ClusterState getClusterStateFromFlinkAPI(FlinkCluster flinkCluster) {
+    final String address = flinkCluster.getAddress();
+    final String jobManagerUrl = flinkCluster.getJobManagerUrl();
+    if (StringUtils.isEmpty(address)) {
+      return ClusterState.STOPPED;
+    }
+    final String flinkUrl =
+        StringUtils.isEmpty(jobManagerUrl)
+            ? address.concat("/overview")
+            : jobManagerUrl.concat("/overview");
```
RocMarshal: After JM failover in Yarn session mode, the `jobManagerUrl` here will change, and the `address` is filled in based on Yarn rest, and will not change after failover.

So could we re-evaluate whether it is really necessary to introduce jobManagerUrl?

RocMarshal (commented):
Hi, @xujiangfeng001 sorry for the late reply due to the necessary peripheral knowledge preparation for the review.
Thanks a lot for the contribution & quick-update.
I left a series of comments about the patch. Please let me know what's your opinion about its.
Thank you.

streampark-common/src/main/java/org/apache/streampark/common/enums/ClusterState.java
xujiangfeng001: When coding, I considered redefining an enumeration class to represent the obtained yarn final status, but later found that adding an enumeration class had redundancy in the status of the platform side. Therefore, I defined the status information of yarn final status in the platform side. Because I think this is non conflicting, and the platform side also needs these states to represent the running status of the flink cluster.

xujiangfeng001: > The two enum look like redundant on platform side. Its could be used to represent the final status on yarn just platform side.

When coding, I considered redefining an enumeration class to represent the obtained yarn final status, but later found that adding an enumeration class had redundancy in the status of the platform side. Therefore, I defined the status information of yarn final status in the platform side. Because I think this is non conflicting, and the platform side also needs these states to represent the running status of the flink cluster.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/FlinkClusterServiceImpl.java
xujiangfeng001: Here, I set the jobmanagerurl to retrieve the status of the flink cluster from the jobmanager. If you obtain the status from the resourcemanager every time, there may be too much pressure on the
resourcemanager when there are too many flink clusters, resulting in a failure to obtain the status. Previously, the status of jobs was obtained from the resourcemanager, which may lead to excessive pressure in the resourcemanager and result in a failure to obtain the status.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkClusterWatcher.java
xujiangfeng001: > AnchorA: What about the Long.MIN_VALUE on init-value ?

It looks very good, I will modify the default values later.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkClusterWatcher.java
xujiangfeng001: > It could be adjusted due to `anchorA`.

I will modify the default values later.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkClusterWatcher.java
xujiangfeng001: > What about ` private static final Duration WATCHER_INTERVAL = Duration.ofSeconds(30);` ?

It looks very good, I will modify the default values later.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkClusterWatcher.java
xujiangfeng001: > The expected state of some clusters is either stopped or abnormal, which should no longer require monitoring. What do you think about on it ?
> 
> Based on the above questions, imagine a situation where Yarn session1 cluster is in a running state. After running for a period of time, the cluster state is FAILED, while Yarn session2 is a cluster that has been created but not started. However, at this point, the platform restarts. How can we ensure that after the restart, only Yarn session1 continues to be monitored and Yarn session2 is ignored?
> 
> For example, introducing an expectState field to identify the expected state of the cluster, only monitoring the heartbeat of clusters with an expected state of RUNNING.
> 
> I'd like to hear anything better about solving the cases.

It looks very good, I will modify the default values later.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkClusterWatcher.java
xujiangfeng001: > What about splitting its into three methods mapping to the corresponding three cases `remote`,`y-session`, `k8s-session` ?

Here, I will reconsider whether to separate or handle it uniformly. I will give you a response after I have considered it thoroughly.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkClusterWatcher.java
xujiangfeng001: > The Yarn resourcemanager rest API will include special cases where the state is accepted but the final status is UNDEFINED.
> 
> So it would be better to judge the current Yarn session status based on `state` and `finalStatus` of yarnAPP.

Here, I will reconsider the processing logic.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkClusterWatcher.java
xujiangfeng001: > After JM failover in Yarn session mode, the `jobManagerUrl` here will change, and the `address` is filled in based on Yarn rest, and will not change after failover.
> 
> So could we re-evaluate whether it is really necessary to introduce jobManagerUrl?

https://github.com/apache/incubator-streampark/pull/2541#discussion_r1164328504
This is why the `jobmanageUrl` was introduced. But regarding the issue of JM's failover, I will reconsider its feasibility.

xujiangfeng001:
> Hi, @xujiangfeng001 sorry for the late reply due to the necessary peripheral knowledge preparation for the review. Thanks a lot for the contribution & quick-update. I left a series of comments about the patch. Please let me know what's your opinion about its. Thank you.

Hi, @RocMarshal thank you very much for your review. I have already responded to the comments. Look forward to your reply.

streampark-common/src/main/java/org/apache/streampark/common/enums/ClusterState.java
RocMarshal: Sorry for the  not clear statements.
I want to express  that platform just focus the status about flink clusters abstraction of platform and should ignore the real status of finalStatus on yarn side. or in other words, the finalStatus is hidden for the Enum.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/FlinkClusterServiceImpl.java
RocMarshal: got it .

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkClusterWatcher.java
RocMarshal: Thx for the clarify. 
got it.

RocMarshal (commented):
hi, @xujiangfeng001  Thank you for you reply.
Would you mind rebasing the latest `dev` branch on your current pr branch as follows?
Because it's recommended to use `rebase` instead of `merge` to sync the latest base branch on our working branch.

- run `git checkout dev` & `git pull origin dev`
- then run `git checkout ISSUE-2498`
- after above,  `git rebase dev`
Maybe it will encounter conflicts.
- finally, squish the muti-commits into one commit record. which is good for reviewers to check it on this a small amount change lines.

Of course, I have to admit that the review progress is not blocked by the mentioned above.
Again, thanks for you patience & effort. 
Looking your next update about the un-addressed comments.

PR #2564, opened 2023-04-04
[FIx] correct the path of shims jar

zhoulii:

the path of shims jar is incorrect, when release app, we cannot pack shims jar into flink-job jar. 

## Verifying this change


wolfboys (approved):
LGTM

PR #2567, opened 2023-04-04
[FIx] rebuild app when flink version has changed

zhoulii:

when flink version has changed, we should rebuild the application, Otherwise, the shims jar packed into flinkjob may be not suitable for the new flink version.

## Verifying this change


wolfboys (approved):
LGTM

PR #2568, opened 2023-04-04
[Style]: Unified Tag Component Style

wangsizhu0504:
## Unified Tag Component Style, Overriding Antd Default Tag Component Style

CleanShot 2023-04-04 at 18 48 21@2x


wolfboys (approved):
LGTM

PR #2583, opened 2023-04-07
[Improve] move jackson setting into profile

VampireAchao:


wolfboys:
Thanks for your contribution。 We need to remove the annotation: `@JsonFormat` on the attribute of beans

VampireAchao:
Okay，I will finish it.

wolfboys (approved):
LGTM

PR #2588, opened 2023-04-08
[Fix] parse key with fraction suffix to float

zhoulii:

the following options should be parsed as float：
- `jobmanager.memory.jvm-overhead.fraction`
- `taskmanager.memory.managed.fraction`
- `taskmanager.memory.jvm-overhead.fraction`

## Verifying this change


wolfboys (approved):
LGTM

PR #2644, opened 2023-04-17
Add @AllowConcurrentEvents annotation

### LINKED ISSUE #2620: [Improve] About the use optimization of AsyncEventBus
### Search before asking

### Description

I found that although asyncEventBus was created in `ChangeEventBus`, the thread pool parameter setting is relatively high, but the subscriber lacks the `@AllowConcurrentEvents` annotation, which will cause asynchronous multi-threading to be ineffective.
```scala
class ChangeEventBus {

  private val execPool = new ThreadPoolExecutor(
    Runtime.getRuntime.availableProcessors * 10,
    Runtime.getRuntime.availableProcessors * 20,
    60L,
    TimeUnit.SECONDS,
    new LinkedBlockingQueue[Runnable](2048))

  private[kubernetes] val asyncEventBus = new AsyncEventBus("[StreamPark][flink-k8s]AsyncEventBus", execPool)

  private[kubernetes] val syncEventBus = new EventBus("[StreamPark][flink-k8s]SyncEventBus")
}
```
In `com.google.common.eventbus.Subscriber`  Subscriber is returned if there is @AllowConcurrentEvents concern on the subscriber method, otherwise SynchronizedSubscriber is returned.
```scala
static Subscriber create(EventBus bus, Object listener, Method method) {
        return (Subscriber)(isDeclaredThreadSafe(method) ? new Subscriber(bus, listener, method) : new Subscriber.SynchronizedSubscriber(bus, listener, method));
    }

private static boolean isDeclaredThreadSafe(Method method) {
        return method.getAnnotation(AllowConcurrentEvents.class) != null;
    }
```
I made a simple example to verify the effect of this annotation [EventBusTest](https://github.com/senlizishi/EventBusTest) . As a result, as mentioned above, no `@AllowConcurrentEvents` is created for synchronous subscribers, and there is no asynchronous effect.

Asynchronous if required. The way to modify is to add `@AllowConcurrentEvents` to the method that needs asynchronous. At the same time, adjust the thread pool parameters to make it more reasonable.
```scala
 private val execPool = new ThreadPoolExecutor(
    Runtime.getRuntime.availableProcessors * 2,
    Runtime.getRuntime.availableProcessors * 3,
    60L,
    TimeUnit.SECONDS,
    new LinkedBlockingQueue[Runnable](1024))
```
The methods I have found to receive AsyncEvent are `FlinkK8sChangeEventListener.subscribeMetricsChange`, `FlinkK8sChangeEventListener.subscribeCheckpointChange`, `BuildInEventListener.subscribeFlinkJobStateEvent`. This requires further discussion and confirmation.

### Usage Scenario

_No response_

### Related issues

_No response_

### Code of Conduct

senlizishi:

Issue Number: close #2620  


senlizishi:
PTAL @Al-assad .  There is no problem in my local test, but changing from synchronous to asynchronous multi-threading may still have many unexposed unknown problems, and more people need to experiment. I still have a question, why does FlinkJobStatusChangeEvent use postAsync in DefaultFlinkK8sWatcher, and FlinkJobStatusChangeEvent uses postSync in FlinkJobStatusWatcher, have you ever encountered such a scenario?

streampark-flink/streampark-flink-kubernetes/src/main/scala/org/apache/streampark/flink/kubernetes/ChangeEventBus.scala
```diff
@@ -24,11 +24,11 @@ import java.util.concurrent.{LinkedBlockingQueue, ThreadPoolExecutor, TimeUnit}
 class ChangeEventBus {
 
   private val execPool = new ThreadPoolExecutor(
-    Runtime.getRuntime.availableProcessors * 10,
-    Runtime.getRuntime.availableProcessors * 20,
+    Runtime.getRuntime.availableProcessors * 2,
```
Al-assad: A small question for me, why is the CPU and buffer queue configuration of the entire ChangeEventBus thread pool narrowed down here?

senlizishi: I think the efficiency has been greatly improved by turning it into multi-threaded asynchronous, so there is no need for such high configuration.

lvshaokang: I think maybe we need a proof that it actually efficient when deploy more jobs. What do you think about it? @Al-assad

Al-assad: Thanks for your reply, as far as I know, most of the subscription programs in this part are blocking IO, and Java threads don't really let out CPU time when IO is blocked like fiber, so it's necessary to set aside a larger thread pool configuration.

Al-assad:
> 
This is determined by the nature of downstream subscription programs. One reason is that postSync downstream can propagate exceptions to upstream calls, while postAsync cannot. This maintains a common exception propagation between upstream and downstream. DefaultFlinkK8sWatcher does not want its own exceptions to be propagated to upstream calls.

Al-assad:
> PTAL @Al-assad . There is no problem in my local test, but changing from synchronous to asynchronous multi-threading may still have many unexposed unknown problems, and more people need to experiment. I still have a question, why does FlinkJobStatusChangeEvent use postAsync in DefaultFlinkK8sWatcher, and FlinkJobStatusChangeEvent uses postSync in FlinkJobStatusWatcher, have you ever encountered such a scenario?

This is determined by the nature of downstream subscription programs. One reason is that postSync downstream can propagate exceptions to upstream calls, while postAsync cannot. This maintains a common exception propagation between upstream and downstream. DefaultFlinkK8sWatcher does not want its own exceptions to be propagated to upstream calls

streampark-flink/streampark-flink-kubernetes/src/main/scala/org/apache/streampark/flink/kubernetes/ChangeEventBus.scala
senlizishi: Thank you for your replies. I will try to conduct a detailed test and give the test results. If you are convenient, you can also do some tests, so that the results can be used for some comparisons to make the results more accurate.

senlizishi: > most of the subscription programs in this part are blocking IO

Hi @Al-assad  I don't quite understand this sentence..  I don't know if your understanding of blocking is consistent with my understanding. My understanding is that after adding `@AllowConcurrentEvents` this annotation will create `Subscriber` instead of `SynchronizedSubscriber`.

In `SynchronizedSubscriber` it is because of the use of `synchronize` that the blocking effect is achieved.
```
@Override
  void invokeSubscriberMethod(Object event) throws InvocationTargetException {
    synchronized (this) {
      super.invokeSubscriberMethod(event);
    }
  }
```

senlizishi: > A small question for me, why is the CPU and buffer queue configuration of the entire ChangeEventBus thread pool narrowed down here?

Hi @Al-assad  . Let me add my understanding on this point. `Runtime.getRuntime().availableProcessors()` gets the number of CPU core threads.  If the number of enabled threads is much greater than the number of CPU core threads, the cost of context switching will be increased, which will affect performance instead.

Not sure if my understanding is correct. looking forward to your reply  :）

Al-assad: Regarding this, you can refer to the formula put forward by Brian Goetz:

```
Number of threads = Number of Available Cores * Target CPU utilization * (1 + Wait time / Service time)
```


For I/O-intensive computations, the blocking coefficient is typically between 0.8 and 0.9. This is a rule of thumb, so the result is approximately: 

```
Number of threads = availableProcessors * 10
Number of threads = availableProcessors * 5
```


Refer：[https://engineering.zalando.com/posts/2019/04/how-to-set-an-ideal-thread-pool-size.html](https://engineering.zalando.com/posts/2019/04/how-to-set-an-ideal-thread-pool-size.html)

senlizishi: @Al-assad  Thanks for the information you provided, I have modified the thread pool parameters. There is also a discussion about blocking IO here.:eyes:

Al-assad (approved):
LGTM. 😃

PR #2650, opened 2023-04-18
[Improve] Bump Hadoop Version to 3.3.4

zhoulii:

there are lots of vulnerabilities in current hadoop dependency, we should bump hadoop to a higher version. this pr upgrades hadoop version to 3.3.4, and replace all hadoop dependencies with `hadoop-client-api` and `hadoop-client-runtime`, and with these changes, we can reduce the number of dependencies in $STREAMPARK/lib from 303 to 242.

this pr was tested on `hadoop-2.10.2` & `hadoop 3.3.5 (the newest version)` with the following operations:

- yarn-application
  - submit application
  - trigger savepoint
  - cancel with savepoint
- yarn-session
  - submit application
  - trigger savepoint
  - cancel with savepoint

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -100,7 +100,7 @@
         <spark.version>3.2.0</spark.version>
         <scalikejdbc.version>4.0.0</scalikejdbc.version>
         <hive.version>2.3.4</hive.version>
-        <hadoop.version>2.10.2</hadoop.version>
+        <hadoop.version>3.3.4</hadoop.version>
         <hbase.version>2.1.10</hbase.version>
         <redis.version>3.3.0</redis.version>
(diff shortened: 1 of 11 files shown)


wolfboys:
Thanks for your contribution. Are other versions of Hadoop compatible?, For example, versions 2. x and 3. x and more

zhoulii:
> 

@wolfboys Thanks for the review. For now Tested with `hadoop-2.10.2`  and  `hadoop 3.3.5` . We should do more tests, but there are too many versions between 2.x and 3.x（about 50+）, maybe we can pick up some typical versions, any suggestions ?

wolfboys (approved):
LGTM

PR #2652, opened 2023-04-18
[Bug] savepoint timeout bug fixed.

wolfboys:
[Bug] savepoint timeout bug fixed.

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/AppBuildPipeServiceImpl.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/AppBuildPipeServiceImpl.java
@@ -26,6 +26,7 @@ import org.apache.streampark.common.fs.FsOperator;
 import org.apache.streampark.common.util.FileUtils;
 import org.apache.streampark.common.util.ThreadUtils;
 import org.apache.streampark.common.util.Utils;
+import org.apache.streampark.console.base.exception.ApiAlertException;
 import org.apache.streampark.console.base.util.WebUtils;
 import org.apache.streampark.console.core.entity.AppBuildPipeline;
 import org.apache.streampark.console.core.entity.Application;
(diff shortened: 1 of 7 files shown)


streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/AppBuildPipeServiceImpl.java
```diff
@@ -210,13 +211,18 @@ public void onStart(PipeSnapshot snapshot) {
...
-                  String uploadJar = localUploads.concat("/").concat(jar);
-                  checkOrElseUploadJar(FsOperator.lfs(), localJar, uploadJar, localUploads);
+                  File uploadJar = new File(localUploads, jar);
+                  if (!localJar.exists() && !uploadJar.exists()) {
+                    throw new ApiAlertException("Missing file: " + jar + ", please upload again");
+                  }
+                  if (localJar.exists()) {
+                    checkOrElseUploadJar(
+                        FsOperator.lfs(), localJar, uploadJar.getAbsolutePath(), localUploads);
+                  }
```
zhoulii, line 225: Do we need to deal with `!localJar.exists() && uploadJar.exists()` ? This may happen when upgrading.

streampark-flink/streampark-flink-client/streampark-flink-client-core/src/main/scala/org/apache/streampark/flink/client/trait/KubernetesNativeClientTrait.scala
```diff
@@ -19,7 +19,6 @@ package org.apache.streampark.flink.client.`trait`
 
 import org.apache.streampark.common.enums.{ExecutionMode, FlinkK8sRestExposedType}
 import org.apache.streampark.flink.client.bean._
-import org.apache.streampark.flink.kubernetes.IngressController
```
zhoulii, line 22: it seems the scala format plugin does not remove unused imports

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/AppBuildPipeServiceImpl.java
wolfboys: > Do we need to deal with `!localJar.exists() && uploadJar.exists()` ? This may happen when upgrading.

`localJar` is find jar under `$APP_HOME/temp`,
`Uploadjar` is find jars in `$workspace/upload`, `$workspace/upload` is a safe directory. The files uploaded by the user in history will be recorded and will not be deleted with the version upgrade.

lvshaokang (approved):
LGTM

PR #2656, opened 2023-04-19
[Bug] yarn session cluster startup failed if no yarn-application job was submitted before

zhoulii:

Hdfs storage won't be initialized until the first yarn-application job is released, so, if we use yarn session mode before using yarn application mode, yarn-session startup would fail.

## Verifying this change


wolfboys (approved):
LGTM

PR #2690, opened 2023-04-24
[Fix]: Issue with abnormal distribution of alarm list at high resolution

wangsizhu0504:
The List component has a screen breakpoint error at high resolutions (>=2000px).


lvshaokang (approved):
LGTM

PR #2696, opened 2023-04-25
[Improve] generate and return random password when reset user password

zhoulii:

generate and return random password to solve possible security issue.

en

Zh

## Verifying this change

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/entity/User.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/entity/User.java
@@ -49,6 +49,8 @@ public class User implements Serializable {
 
   public static final String SEX_UNKNOWN = "2";
 
+  public static final Integer DEFAULT_PASSWORD_LENGTH = 8;
+
   public static final String DEFAULT_PASSWORD = "streampark666";
 
   @TableId(type = IdType.AUTO)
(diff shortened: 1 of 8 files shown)


streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/entity/User.java
```diff
@@ -49,6 +49,8 @@ public class User implements Serializable {
 
   public static final String SEX_UNKNOWN = "2";
 
+  public static final Integer DEFAULT_PASSWORD_LENGTH = 8;
+
   public static final String DEFAULT_PASSWORD = "streampark666";
```
wolfboys: delete line 54.

streampark-console/streampark-console-webapp/src/views/system/user/User.vue
```diff
@@ -160,8 +160,14 @@
...
         const hide = createMessage.loading('reseting');
         try {
-          await resetPassword({ usernames: record.username });
-          Swal.fire(t('system.user.table.resetSuccess', [record.username]), '', 'success');
+          const resp = await resetPassword({ username: record.username });
+          if (resp.data.code == 200) {
+            Swal.fire({
+              icon: 'success',
+              title: 'New Password',
+              html: '<pre class="api-exception"> The new password is: ' + resp.data.data + '</pre>',
```
wolfboys: need to support i18n.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/entity/User.java
zhoulii: done

streampark-console/streampark-console-webapp/src/views/system/user/User.vue
zhoulii: done

wolfboys (approved):
LGTM

Issue #2723, opened 2023-05-05
[Bug] Compatibility problem under fink 1.17.0 version when restoring job from savepoint


cygnusdark:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-streampark/issues?q=is%3Aissue+label%3A%22bug%22) and found no similar issues.

### Java Version

1.8

### Scala Version

2.12.x

### StreamPark Version

2.1.0

### Flink Version

1.17.0

### deploy mode

yarn-application

### What happened

When I restart a job from previous savepoint, I got an error exception.

It looks like a compatibility problem.

It seems like we need to upgrade commons-cli version to 1.5, or upgrade the Flink version in Flink shims base

Is there better way to fix this problem?

### Error Exception

```log
Caused by: java.lang.NoSuchMethodError: org.apache.commons.cli.CommandLine.hasOption(Lorg/apache/commons/cli/Option;)Z
	at org.apache.flink.client.cli.CliFrontendParser.createSavepointRestoreSettings(CliFrontendParser.java:631)
	at org.apache.flink.client.cli.ProgramOptions.<init>(ProgramOptions.java:119)
	at org.apache.flink.client.cli.ProgramOptions.create(ProgramOptions.java:192)
	at org.apache.streampark.flink.client.trait.FlinkClientTrait.submit(FlinkClientTrait.scala:87)
	at org.apache.streampark.flink.client.trait.FlinkClientTrait.submit$(FlinkClientTrait.scala:63)
	at org.apache.streampark.flink.client.impl.YarnApplicationClient$.submit(YarnApplicationClient.scala:44)
	at org.apache.streampark.flink.client.FlinkClientHandler$.submit(FlinkClientHandler.scala:40)
	at org.apache.streampark.flink.client.FlinkClientHandler.submit(FlinkClientHandler.scala)
```

### Screenshots

<img width="1581" alt="image" src="https://user-images.githubusercontent.com/53288632/236369661-f06cf84d-bc10-461a-bbbb-ae1db36ea429.png">

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!(您是否要贡献这个PR?)

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wolfboys:
Our developer has tested it and found no issues running it under Flink 1.17. You may want to check if there are any commons-cli related jars in your flink/lib directory.

[...]

PR #2725, opened 2023-05-05
[Bugfix]Bump commons-cli version to resolve compatibility problem under flink 1.17

cygnusdark:
Since flink 1.15, the commons cli version used by flink clients package is 1.5.0.
Streampark currently uses commons cli 1.3.1, which is not compatible with flink 1.17.0.
See https://github.com/apache/incubator-streampark/issues/2723
Upgrade the commons cli version will fix this problem.
I have tested this modification in prod environment under flink 1.16.0 & 1.17.0 and streampark 2.1.0-rc1.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -117,7 +117,7 @@
         <snakeyaml.version>2.0</snakeyaml.version>
         <typesafe-conf.version>1.4.2</typesafe-conf.version>
         <json4s-jackson.version>4.0.6</json4s-jackson.version>
-        <commons-cli.version>1.3.1</commons-cli.version>
+        <commons-cli.version>1.5</commons-cli.version>
         <commons-net.version>3.9.0</commons-net.version>
         <commons-lang3.version>3.8.1</commons-lang3.version>
         <enumeratum.version>1.6.1</enumeratum.version>
--- a/pom.xml
+++ b/pom.xml
@@ -117,7 +117,7 @@
         <snakeyaml.version>2.0</snakeyaml.version>
         <typesafe-conf.version>1.4.2</typesafe-conf.version>
         <json4s-jackson.version>4.0.6</json4s-jackson.version>
-        <commons-cli.version>1.5</commons-cli.version>
+        <commons-cli.version>1.5.0</commons-cli.version>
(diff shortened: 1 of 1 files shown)


zhoulii:
Hi @cygnusdark , Thanks for your contribution. We tested what's you report and cannot reproduce this issue. Actually, StreamPark uses proxy classloader to submit flink job, whick means the classes are loaded from `$FLINK_HOME/lib`, It's highly possible that some dependency in your `$FLINK_HOME/lib` conflict with flink-dist.

Though we cannot reproduce this problem, this pr that bumps commons-cli version to 1.5.0 looks good to me.

wolfboys:
> Hi @cygnusdark , Thanks for your contribution. We tested what's you report and cannot reproduce this issue. Actually, StreamPark uses proxy classloader to submit flink job, whick means the classes are loaded from `$FLINK_HOME/lib`, It's highly possible that some dependency in your `$FLINK_HOME/lib` conflict with flink-dist.
> 
> Though we cannot reproduce this problem, this pr that bumps commons-cli version to 1.5.0 looks good to me.

We can expand the scope of testing, for example from Flink 1.12 to 1.17, to further verify compatibility. If the tests pass, I would be happy to merge this pull request.

cygnusdark:
@wolfboys Thank you for your support😊. I am still looking into the possible dependency conflict @zhoulii mentioned. The solution that bumping the commons-cli version has already fixed my problem. If I find the root cause later, I will share here.

cygnusdark:
Recently, another streampark user report same error：[Bug] NoSuchMethodError when starting job #2803.
I think this problem happens when users use hadoop 2.x jars, as this article mentioned: https://stackoverflow.com/questions/38654957/hadoop-nosuchmethoderror-apache-commons-cli

Compared with modifying hadoop jar, this pr is a easier way for some users. @wolfboys Could you help merge this pr？

wolfboys:
Recently, Some Streampark users have also encountered this issue. After my investigation, I found out that it is not simply caused by dependency conflicts. These few days, I will verify this issue and test current pull request. If it works ok, I am very willing to merge it.

wolfboys (approved):
LGTM

PR #2733, opened 2023-05-07
[improve] Upgrade httpclient version to 5.1

wolfboys:
[improve] Upgrade httpclient version to 5.1

### DIFF
--- a/streampark-common/src/main/scala/org/apache/streampark/common/util/HttpClientUtils.scala
+++ b/streampark-common/src/main/scala/org/apache/streampark/common/util/HttpClientUtils.scala
@@ -16,23 +16,23 @@
  */
 package org.apache.streampark.common.util
 
-import org.apache.http.NameValuePair
-import org.apache.http.auth.{AuthSchemeProvider, AuthScope, Credentials}
-import org.apache.http.client.config.{AuthSchemes, RequestConfig}
-import org.apache.http.client.entity.UrlEncodedFormEntity
-import org.apache.http.client.methods._
-import org.apache.http.client.utils.URIBuilder
-import org.apache.http.config.RegistryBuilder
-import org.apache.http.entity.StringEntity
-import org.apache.http.impl.auth.SPNegoSchemeFactory
-import org.apache.http.impl.client.{BasicCredentialsProvider, CloseableHttpClient, HttpClientBuilder, HttpClients}
-import org.apache.http.impl.conn.PoolingHttpClientConnectionManager
-import org.apache.http.message.BasicNameValuePair
-import org.apache.http.util.EntityUtils
+import org.apache.hadoop.shaded.org.apache.http.client.config.AuthSchemes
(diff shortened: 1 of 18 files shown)


streampark-common/src/main/scala/org/apache/streampark/common/util/HttpClientUtils.scala
```diff
@@ -16,23 +16,23 @@
...
-import org.apache.http.client.methods._
-import org.apache.http.client.utils.URIBuilder
-import org.apache.http.config.RegistryBuilder
-import org.apache.http.entity.StringEntity
-import org.apache.http.impl.auth.SPNegoSchemeFactory
-import org.apache.http.impl.client.{BasicCredentialsProvider, CloseableHttpClient, HttpClientBuilder, HttpClients}
-import org.apache.http.impl.conn.PoolingHttpClientConnectionManager
-import org.apache.http.message.BasicNameValuePair
-import org.apache.http.util.EntityUtils
+import org.apache.hadoop.shaded.org.apache.http.client.config.AuthSchemes
```
zhoulii: We'd better not use reference from `hadoop.shaded` package.

streampark-common/src/main/scala/org/apache/streampark/common/util/HttpClientUtils.scala
```diff
@@ -97,83 +98,47 @@ object HttpClientUtils {
...
+        def getUserPrincipal: Principal = null
+      }
+
+      credentialsProvider.setCredentials(new AuthScope(null, -1), credentials)
 
       val authSchemeRegistry = RegistryBuilder
-        .create[AuthSchemeProvider]
-        .register(AuthSchemes.SPNEGO, new SPNegoSchemeFactory(true))
+        .create[AuthSchemeFactory]
+        .register(AuthSchemes.SPNEGO, SPNegoSchemeFactory.DEFAULT)
```
zhoulii: `AuthSchemes` is removed in 5.1, we can use `StandardAuthScheme` instead.

zhoulii (commented):
Nice work, I left a minor comment, please take a look.

streampark-common/src/main/scala/org/apache/streampark/common/util/HttpClientUtils.scala
wolfboys: > `AuthSchemes` is removed in 5.1, we can use `StandardAuthScheme` instead.

Thanks for your review, I'll update it later

zhoulii:
Deployed and tested with yarn `application/session/perjob` execution mode, no bug was found, so LGTM.

wolfboys:
> Deployed and tested with yarn `application/session/perjob` execution mode, no bug was found, so LGTM.

thanks for your review and test.

wangsizhu0504 (approved):
LGTM

PR #2740, opened 2023-05-11
[FIX] enable token expire time

zhoulii:

enable token expire time which is unchangeable now.

token-expire-time-is-disabled

### DIFF
--- a/streampark-console/streampark-console-webapp/src/locales/lang/en/flink/project.ts
+++ b/streampark-console/streampark-console-webapp/src/locales/lang/en/flink/project.ts
@@ -50,7 +50,7 @@ export default {
     branchesPlaceholder: 'Select a branch',
     pom: 'POM',
     pomPlaceholder:
-      'By default,lookup pom.xml in root path,You can manually specify the module to compile pom.xmlh',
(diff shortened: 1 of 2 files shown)


GOODBOY008:
It seems like disable expire time function manually. So I think if we confirm this function is useless and delete this function. @wolfboys @zhoulii cc.

zhoulii:
> It seems like disable expire time function manually. So I think if we confirm this function is useless and delete this function. @wolfboys @zhoulii cc.

In my opinion, support token expire time is needed, I'm not sure why this function is disabled, maybe by accident ? 😂

GOODBOY008:
> > It seems like disable expire time function manually. So I think if we confirm this function is useless and delete this function. @wolfboys @zhoulii cc.
> 
> In my opinion, support token expire time is needed, I'm not sure why this function is disabled, maybe by accident ? 😂

So we need to dig the reason why disable date selector and set expire time to `9999-01-01` by default. Finally ,we discuss whether this fuction is need.

wolfboys:
set expiry time for tokens is not a good solution. A better way is for admin to be able to manually revoke tokens, which means the admin can make the token expire at any time. We can discuss how to implement this.

wolfboys:
resolved in version 2.1.3

PR #2751, opened 2023-05-16
[Bug] Kubernetes session cluster can be deleted even if it's still in use

zhoulii:

Kubernetes session cluster can be deleted even if it's still in use, that because application which ran in kubernetes session mode does not set flinkClusterId field in database.

## Verifying this change


wolfboys (approved):
LGTM

PR #2793, opened 2023-06-06
[improve] hadoop sample support

wolfboys:
[improve] hadoop sample support

## Contribution Checklist

hadoop sample support, [hadoop Authentication doc](https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-common/HttpAuthentication.html)

### DIFF
--- a/streampark-common/src/main/scala/org/apache/streampark/common/util/HadoopUtils.scala
+++ b/streampark-common/src/main/scala/org/apache/streampark/common/util/HadoopUtils.scala
@@ -58,7 +58,7 @@ object HadoopUtils extends Logger {
 
   private[this] var tgt: KerberosTicket = _
 
-  private lazy val hadoopUserName: String =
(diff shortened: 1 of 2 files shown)


wolfboys:
cc @zhoulii @macksonmu PTAL, thx

zhoulii:
Tested, works well. But It seems, with or without this change, there are no abnormal issues when enable `simple` Authentication.

wolfboys:
> Tested, works well. But It seems, with or without this change, there are no abnormal issues when enable `simple` Authentication.

enable sample authentication in Hadoop generally does not require any special handling. However, recently there has been feedback from users in the community that related requests need to carry the `user.name` parameter to function properly.

muchunjin (commented):
LGTM

muchunjin (approved):
LGTM

PR #2798, opened 2023-06-14
yarn resourceManager URL bug fix

wolfboys:
yarn resourceManager URL bug fix

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/bean/AlertTemplate.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/bean/AlertTemplate.java
@@ -60,7 +60,7 @@ public class AlertTemplate implements Serializable {
 
     if (ExecutionMode.isYarnMode(application.getExecutionMode())) {
       String format = "%s/proxy/%s/";
-      String url = String.format(format, YarnUtils.getRMWebAppURL(), application.getAppId());
+      String url = String.format(format, YarnUtils.getRMWebAppURL(false), application.getAppId());
       template.setLink(url);
     } else {
       template.setLink(null);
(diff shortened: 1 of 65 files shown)


wolfboys:
cc @macksonmu @zhoulii PTAL

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/bean/AlertTemplate.java
```diff
@@ -60,7 +60,7 @@ private static AlertTemplate of(Application application) {
 
     if (ExecutionMode.isYarnMode(application.getExecutionMode())) {
       String format = "%s/proxy/%s/";
-      String url = String.format(format, YarnUtils.getRMWebAppURL(), application.getAppId());
+      String url = String.format(format, YarnUtils.getRMWebAppURL(false), application.getAppId());
```
zhoulii, line 63: do we have to get dynamic url here ? if not, we can omit the false parameter.

zhoulii (commented):
Thanks for your contribution @wolfboys , it looks good to me, just left one minor comment.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/bean/AlertTemplate.java
wolfboys: > do we have to get dynamic url here ? if not, we can omit the false parameter.

If the latest YARN RM url is obtained every time, it will result in more network requests and resource consumption. Obtaining a currently valid active rm url requires a lot of work, as you can see in `org.apache.hadoop.yarn.util.RMHAUtils.findActiveRMHAId` and [streampark](https://github.com/apache/incubator-streampark/blob/827fe32449a25d05a73999a59727f782529e69ff/streampark-common/src/main/scala/org/apache/streampark/common/util/YarnUtils.scala#L159.), by default, we return the previously active rm url. If this url is not available, we will retrieve the latest active node again.

muchunjin (approved):
Thank you for your hard work. We conducted resource manager master-slave switching tests in per job, application, and Yarn session modes, and the test results showed that the job status was normal. After the master-slave switching, the address of the request URL printed in the backend log changed from the old resourcemanager address to the new master resourcemanager address

PR #2809, opened 2023-06-20
[ISSUE-2423][Feature][WIP] flink cluster failure alarm&failover

xujiangfeng001:
## What changes were proposed in this pull request

Issue Number: close https://github.com/apache/incubator-streampark/issues/2423

## Brief change log

For details, please refer to: https://github.com/apache/incubator-streampark/issues/2423#issuecomment-1537068906.

This PR only completes the server part, and the client configuration `alert` will be completed later.

## Verifying this change

This change is a trivial rework / code cleanup without any test coverage.

## Does this pull request potentially affect one of the following parts
 - Dependencies (does it add or upgrade a dependency): (no)

PR #2826, opened 2023-07-01
[Improve] Flink cluster status monitoring improvement

wolfboys:
[Improve] Flink cluster status monitoring improvement

### DIFF
--- a/streampark-console/streampark-console-service/src/main/assembly/script/schema/mysql-schema.sql
+++ b/streampark-console/streampark-console-service/src/main/assembly/script/schema/mysql-schema.sql
@@ -438,7 +438,6 @@ drop table if exists `t_flink_cluster`;
 create table `t_flink_cluster` (
   `id` bigint not null auto_increment,
   `address` varchar(150) default null comment 'url address of cluster',
-  `job_manager_url` varchar(150) default null comment 'url address of jobmanager',
   `cluster_id` varchar(45) default null comment 'clusterid of session mode(yarn-session:application-id,k8s-session:cluster-id)',
   `cluster_name` varchar(128) not null comment 'cluster name',
   `options` text comment 'json form of parameter collection ',
--- a/streampark-console/streampark-console-service/src/main/assembly/script/schema/mysql-schema.sql
+++ b/streampark-console/streampark-console-service/src/main/assembly/script/schema/mysql-schema.sql
@@ -438,6 +438,7 @@ drop table if exists `t_flink_cluster`;
 create table `t_flink_cluster` (
   `id` bigint not null auto_increment,
   `address` varchar(150) default null comment 'url address of cluster',
+  `job_manager_url` varchar(150) default null comment 'url address of jobmanager',
   `cluster_id` varchar(45) default null comment 'clusterid of session mode(yarn-session:application-id,k8s-session:cluster-id)',
   `cluster_name` varchar(128) not null comment 'cluster name',
   `options` text comment 'json form of parameter collection ',
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkClusterWatcher.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkClusterWatcher.java
@@ -30,6 +30,7 @@ import org.apache.streampark.console.core.service.ApplicationService;
 import org.apache.streampark.console.core.service.FlinkClusterService;
 import org.apache.streampark.console.core.service.alert.AlertService;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.yarn.api.records.FinalApplicationStatus;
 import org.apache.hc.client5.http.config.RequestConfig;
 
@@ -167,7 +168,14 @@ public class FlinkClusterWatcher {
    */
   private ClusterState httpRemoteClusterState(FlinkCluster flinkCluster) {
     final String address = flinkCluster.getAddress();
-    final String flinkUrl = address.concat("/overview");
+    if (StringUtils.isEmpty(address)) {
+      return ClusterState.STOPPED;
+    }
+    final String jobManagerUrl = flinkCluster.getJobManagerUrl();
+    final String flinkUrl =
+        StringUtils.isEmpty(jobManagerUrl)
+            ? address.concat("/overview")
+            : jobManagerUrl.concat("/overview");
     try {
       String res =
           HttpClientUtils.httpGetRequest(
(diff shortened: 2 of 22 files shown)


wolfboys:
cc @xujiangfeng001 @RocMarshal PTAL, thx

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkClusterWatcher.java
```diff
@@ -321,11 +251,9 @@ private FinalApplicationStatus stringConvertFinalApplicationStatus(String value)
    * @return
    */
   private ClusterState finalApplicationStatusConvertClusterState(FinalApplicationStatus status) {
-    switch (status) {
-      case UNDEFINED:
-        return ClusterState.RUNNING;
-      default:
-        return ClusterState.STOPPED;
+    if (status == FinalApplicationStatus.UNDEFINED) {
```
RocMarshal, line 262: When the yarn application was accepted by yarn resourcemanager, the FinalApplicationStatus is `UNDEFINED` too, but the flink cluster isn't `RUNNING`

This case was mentioned in https://github.com/apache/incubator-streampark/pull/2541#discussion_r1162899796

streampark-console/streampark-console-service/src/main/resources/mapper/core/ApplicationMapper.xml
```diff
@@ -133,7 +133,7 @@
              limit 1
     </select>
 
-    <select id="getAffectedJobsByClusterId" resultType="java.lang.Integer" parameterType="java.lang.Long">
+    <select id="countJobsByClusterId" resultType="java.lang.Integer" parameterType="java.lang.Long">
         select
             count(1)
         from t_flink_app
```
RocMarshal, line 139: Do we need to filter the status of the application here? https://github.com/apache/incubator-streampark/pull/2809#discussion_r1242367682

RocMarshal (commented):
Thanks for @wolfboys @xujiangfeng001 driving this pr.
I left a few of comments, PTAL in your free time.
Thank you~

streampark-console/streampark-console-service/src/main/resources/mapper/core/ApplicationMapper.xml
wolfboys: > Do we need to filter the status of the application here? [#2809 (comment)](https://github.com/apache/incubator-streampark/pull/2809#discussion_r1242367682)

Thanks for your feedback. I agree with you. I think We should filter the status of the application

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkClusterWatcher.java
wolfboys: > When the yarn application was accepted by yarn resourcemanager, the FinalApplicationStatus is `UNDEFINED` too, but the flink cluster isn't `RUNNING`
> 
> This case was mentioned in [#2541 (comment)](https://github.com/apache/incubator-streampark/pull/2541#discussion_r1162899796)

I also have questions about this logic and further discussion is needed. @xujiangfeng001

streampark-console/streampark-console-service/src/main/assembly/script/schema/mysql-schema.sql
```diff
@@ -438,7 +438,6 @@ drop table if exists `t_flink_cluster`;
 create table `t_flink_cluster` (
   `id` bigint not null auto_increment,
   `address` varchar(150) default null comment 'url address of cluster',
-  `job_manager_url` varchar(150) default null comment 'url address of jobmanager',
```
xujiangfeng001: We need to confirm if we still need a ` job_ manager_ url`  to alleviate the request pressure of `resourceManager`.

streampark-console/streampark-console-service/src/main/resources/mapper/core/ApplicationMapper.xml
xujiangfeng001: > > Do we need to filter the status of the application here? [#2809 (comment)](https://github.com/apache/incubator-streampark/pull/2809#discussion_r1242367682)
> 
> Thanks for your feedback. I agree with you. I think We should filter the status of the application

It is difficult to directly query the data table for this issue. I think we need a cache to record the `running jobs` of each `flink cluster`.

xujiangfeng001:
Hi @wolfboys @RocMarshal  ,I left a few of comments. Please take a look at it when you have time.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkClusterWatcher.java
xujiangfeng001: This is my negligence. The corresponding logic was not modified during the implementation of this piece of content. The logic here does not need to determine the `finalStatus `of the application, only the `application state` needs to be determined as `running` to consider the `flink cluster` as `running`.

```
public enum YarnApplicationState {
  /** Application which was just created. */
  NEW,

  /** Application which is being saved. */
  NEW_SAVING,

  /** Application which has been submitted. */
  SUBMITTED,

  /** Application has been accepted by the scheduler */
  ACCEPTED,

  /** Application which is currently running. */
  RUNNING,

  /** Application which finished successfully. */
  FINISHED,

  /** Application which failed. */
  FAILED,

  /** Application which was terminated by a user or admin. */
  KILLED
}
```

streampark-console/streampark-console-service/src/main/assembly/script/schema/mysql-schema.sql
wolfboys: > We need to confirm if we still need a ` job_ manager_ url` to alleviate the request pressure of `resourceManager`.

done

wolfboys:
cc @RocMarshal @xujiangfeng001  PTAL

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkClusterWatcher.java
```diff
@@ -168,13 +135,29 @@ private void alert(FlinkCluster cluster, ClusterState state) {
    * @param flinkCluster
    * @return
    */
-  private ClusterState getClusterState(FlinkCluster flinkCluster) {
-    ClusterState state = getClusterStateFromFlinkAPI(flinkCluster);
-    if (ClusterState.isRunningState(state)) {
+  public ClusterState getClusterState(FlinkCluster flinkCluster) {
+    ClusterState state = FAILED_STATES.getIfPresent(flinkCluster.getId());
```
xujiangfeng001, line 139: I think there is a problem with the logic in this method: here, in Yarn Session mode, it won't go `jobmanagerUrl `request status, only `yarn restful api` requests will be processed. This is not in line with our original intention of designing the `jobmanagerUrl` field.

xujiangfeng001:
Hi @wolfboys ，this generally looks good, I left a comment. If I make a mistake, please correct me.

RocMarshal (approved):
Thanks @wolfboys & @xujiangfeng001 .
LGTM +1.

PR #2850, opened 2023-07-12
[Fix]: monaco editor foldingStrategy parameters

wangsizhu0504:
## Fix the problem of invalid foldingStrategy in SQL editor parameter

iShot_2023-07-12_11 55 52

Editor adds code folding


wolfboys (approved):
LGTM

Issue #2865, opened 2023-07-18
[improvement] improvement flink cluster alert


xujiangfeng001:
### Search before asking

- [X] I had searched in the [feature](https://github.com/apache/incubator-streampark/issues?q=is%3Aissue+label%3A%22Feature%22) and found no similar feature requirement.

### Description

1. add `total number of jobs` information to the alarm
2. optimize the calculation of `affecting the number of jobs` in alarms 
3. add `flink cluster state` display on the front end

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #2875, opened 2023-07-21
[improvement] improvement flink cluster alert

xujiangfeng001:

Issue Number: https://github.com/apache/incubator-streampark/issues/2865

## Brief change log

1. add `total number of jobs` information to the alarm

2. optimize the calculation of `affecting the number of jobs` in alarms

3. add `flink cluster state` display on the front end

## Verifying this change


xujiangfeng001:
cc @wolfboys @RocMarshal @wangsizhu0504  , PTAL,thx.

streampark-console/streampark-console-service/src/main/resources/mapper/core/ApplicationMapper.xml
```diff
@@ -141,6 +141,16 @@
             limit 1
     </select>
 
+    <select id="countAffectedJobsByClusterId" resultType="java.lang.Integer" parameterType="java.lang.Long">
+        select
+            count(1)
+        from t_flink_app
+        where flink_cluster_id = #{clusterId}
+            and state in (5, 7, 9)
+            and (end_time IS NULL or TIMESTAMPDIFF(SECOND, end_time, NOW()) &lt;= 1)
```
wolfboys: Does TIMESTAMPDIFF support in h2 and pgsql ?

xujiangfeng001: Thank you very much for the review. I have made changes to the code. Could you please review it when you have time.

xujiangfeng001: > Does TIMESTAMPDIFF support in h2 and pgsql ?

Hi @wolfboys ，Thank you very much for the review. I have made changes to the code. Could you please review it when you have time.

streampark-console/streampark-console-webapp/src/enums/flinkEnum.ts
```diff
@@ -140,11 +140,17 @@ export enum ClusterStateEnum {
...
   /** cluster started */
-  STARTED = 1,
+  RUNNING = 1,
   /** cluster canceled */
   CANCELED = 2,
   /** cluster lost */
   LOST = 3,
+  /** cluster unknown */
+  UNKNOWN = 4,
+  STARTING = 5,
```
wolfboys, line 150: These values are too arbitrary.  We can set the status according to certain rules, such as transitioning from "create" to "start" to "stop." and then list some status of each stage, which would be better

streampark-common/src/main/java/org/apache/streampark/common/enums/ClusterState.java
```diff
@@ -32,7 +32,7 @@ public enum ClusterState implements Serializable {
   /** cluster unknown */
   UNKNOWN(4),
   STARTING(5),
-  CANCELING(6),
+  CANCELLING(6),
```
wolfboys, line 35: These values are too arbitrary. We can set the status according to certain rules, such as from "create" to "start" to "stop", and then list some status of each stage,  e.g:
```  
CREATED(0),
STARTING(1),
RUNNING(2),
CANCELLING(3),
CANCELED(4),
FAILED(5),
KILLED(6);
LOST(7),
UNKNOWN(8)
```

xujiangfeng001: Due to historical status reasons, directly modifying the enumeration value may cause confusion in the historical status of the `flink cluster`. So I kept the original enumeration values.

wolfboys: > Due to historical status reasons, directly modifying the enumeration value may cause confusion in the historical status of the `flink cluster`. So I kept the original enumeration values.

thanks for the clarification

wolfboys (approved):
LGTM

PR #2912, opened 2023-07-30
[optimization] console service code optimization

VampireAchao:

Issue Number: close #xxx 

## Verifying this change

*(or)*

*(or)*

 no

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/domain/RestResponse.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/domain/RestResponse.java
@@ -17,6 +17,8 @@
 
 package org.apache.streampark.console.base.domain;
 
+import org.jetbrains.annotations.NotNull;
+
 import java.util.HashMap;
 
 public class RestResponse extends HashMap<String, Object> {
@@ -60,6 +62,7 @@ public class RestResponse extends HashMap<String, Object> {
     return this;
   }
 
+  @NotNull
   @Override
   public RestResponse put(String key, Object value) {
     super.put(key, value);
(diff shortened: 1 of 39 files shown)


streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/domain/RestResponse.java
```diff
@@ -17,6 +17,8 @@
 
 package org.apache.streampark.console.base.domain;
 
+import org.jetbrains.annotations.NotNull;
```
wolfboys: this dependency (maven pom) is not imported in any pom, in other words, can it still be compiled in a non intellij idae?

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/util/CommonUtils.java
```diff
@@ -93,7 +90,8 @@ public static Boolean isEmpty(Object... objs) {
       }
 
       if (obj instanceof Iterable) {
-        if (((Iterable<?>) obj).iterator() == null || !((Iterable<?>) obj).iterator().hasNext()) {
+        ((Iterable<?>) obj).iterator();
```
wolfboys: If obj is null, will an exception occur?

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/util/CommonUtils.java
```diff
@@ -191,11 +189,11 @@ public static Float toFloat(Object val) {
     return toFloat(val, 0f);
   }
 
-  public static List arrayToList(Object source) {
+  public static List<?> arrayToList(Object source) {
     return Arrays.asList(ObjectUtils.toObjectArray(source));
   }
 
-  public static boolean contains(Iterator iterator, Object element) {
+  public static boolean contains(Iterator<Object> iterator, Object element) {
```
wolfboys: I don't think we need to change here

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/util/CommonUtils.java
VampireAchao: it will not

PR #2913, opened 2023-07-31
IngressController code optimization

ChengJie1053:
IngressController code optimization

### DIFF
--- a/streampark-flink/streampark-flink-kubernetes/src/main/scala/org/apache/streampark/flink/kubernetes/KubernetesRetriever.scala
+++ b/streampark-flink/streampark-flink-kubernetes/src/main/scala/org/apache/streampark/flink/kubernetes/KubernetesRetriever.scala
@@ -47,15 +47,24 @@ object KubernetesRetriever extends Logger {
   // see org.apache.flink.configuration.RestOptions.RETRY_MAX_ATTEMPTS
   val FLINK_REST_RETRY_MAX_ATTEMPTS = 2
 
+  private var kubernetesClient: KubernetesClient = _
+
   /** get new KubernetesClient */
   @throws(classOf[KubernetesClientException])
-  def newK8sClient(): KubernetesClient = {
-    new DefaultKubernetesClient()
+  def getK8sClient(): KubernetesClient = {
+    if (kubernetesClient == null) {
+      synchronized {
+        if (kubernetesClient == null) {
+          kubernetesClient = new DefaultKubernetesClient()
+        }
+      }
+    }
+    kubernetesClient
   }
 
   /** check connection of kubernetes cluster */
   def checkK8sConnection(): Boolean = {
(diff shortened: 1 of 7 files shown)


streampark-flink/streampark-flink-kubernetes/src/main/scala/org/apache/streampark/flink/kubernetes/KubernetesRetriever.scala
```diff
@@ -47,15 +47,24 @@ object KubernetesRetriever extends Logger {
   // see org.apache.flink.configuration.RestOptions.RETRY_MAX_ATTEMPTS
   val FLINK_REST_RETRY_MAX_ATTEMPTS = 2
 
+  private var kubernetesClient: KubernetesClient = _
+
   /** get new KubernetesClient */
   @throws(classOf[KubernetesClientException])
-  def newK8sClient(): KubernetesClient = {
-    new DefaultKubernetesClient()
+  def getK8sClient(): KubernetesClient = {
```
Al-assad, line 54: There are better ways to implement the singleton pattern in scala:

```scala 
private lazy val kubernetesClient: KubernetesClient = new DefaultKubernetesClient()
def getK8sClient(): KubernetesClient = kubernetesClient
```

Al-assad (changes requested):

streampark-flink/streampark-flink-kubernetes/src/main/scala/org/apache/streampark/flink/kubernetes/KubernetesRetriever.scala
```diff
@@ -107,7 +116,7 @@ object KubernetesRetriever extends Logger {
    *   deployment namespace
    */
   def isDeploymentExists(name: String, namespace: String): Boolean = {
-    using(KubernetesRetriever.newK8sClient()) {
+    using(KubernetesRetriever.getK8sClient()) {
```
wolfboys, line 119: The KubernetesRetriever.getK8sClient() method will return a singleton KubernetesClient. However, the using method closes the KubernetesClient after used, which can lead to bugs.

streampark-flink/streampark-flink-kubernetes/src/main/scala/org/apache/streampark/flink/kubernetes/KubernetesRetriever.scala
ChengJie1053: Ok, thanks for reviewing the code

streampark-flink/streampark-flink-kubernetes/src/main/scala/org/apache/streampark/flink/kubernetes/KubernetesRetriever.scala
ChengJie1053: Ok, thanks for reviewing the code

PR #2922, opened 2023-08-04
dependency minor improvement

wolfboys:
dependency minor improvement


wolfboys:
cc @Al-assad PTAL, thx

streampark-common/src/main/scala/org/apache/streampark/common/zio/ZIOLogger.scala
```diff
@@ -73,6 +73,7 @@ object LoggerBackend {
       case LogLevel.Warning => logger.warn(msg)
       case LogLevel.Error => logger.error(msg)
       case LogLevel.Fatal => logger.error(msg)
+      case _ => logger.error(msg)
```
Al-assad: All cases of ZIO.logging.LogLevel are already included here, the remaining matches are not necessary,. If you force a remaining match, it is safer to match to `logger.trace`

Al-assad (approved):
LGTM

PR #2943, opened 2023-08-14
ResourceServiceImpl  code optimization

ChengJie1053:
ResourceServiceImpl  code optimization


streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/ResourceServiceImpl.java
```diff
@@ -284,7 +285,7 @@ public RestResponse checkResource(Resource resourceParam) throws JsonProcessingE
         Manifest manifest = Utils.getJarManifest(jarFile);
         String mainClass = manifest.getMainAttributes().getValue("Main-Class");
 
-        if (mainClass == null) {
+        if (mainClass == null && StringUtils.isBlank(resourceParam.getMainClass())) {
```
wolfboys, line 288: oh no, The main-class must be parseable from the user's flink app, otherwise it is an invalid flink app

ChengJie1053: Ok, thanks for reviewing the code

PR #2946, opened 2023-08-15
docs: drop vanity metrics from README

tisonkun:
These photos are unstable to render that can cause negative reading experience.

Also, the number of stars is shown right upper and contributors are shown on the above contributors big graph.

We don't need too much vanity metrics.

image


wolfboys (approved):
LGTM

PR #2948, opened 2023-08-15
jar upload bug fix

ChengJie1053:
jar upload bug fix

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/entity/Resource.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/entity/Resource.java
@@ -82,4 +82,29 @@ public class Resource implements Serializable {
   private transient String sortOrder;
 
   private transient String connector;
+
+  public void setResourcePath(String resourcePath) {
+    if (resourcePath == null) {
+      throw new IllegalArgumentException("resource path cannot be null.");
+    }
+    String[] namePath = resourcePath.split(":");
+    if (namePath.length != 2) {
+      throw new IllegalArgumentException("resource path invalid, format: $name:$path");
+    }
+    this.resourcePath = resourcePath;
+  }
+
+  public String getFileName() {
+    if (this.resourcePath == null) {
+      return null;
+    }
+    return resourcePath.split(":")[0];
+  }
+
+  public String getFilePath() {
+    if (this.resourcePath == null) {
+      return null;
+    }
+    return resourcePath.split(":")[1];
+  }
 }
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/AppBuildPipeServiceImpl.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/AppBuildPipeServiceImpl.java
@@ -80,6 +80,7 @@ import org.apache.streampark.flink.packer.pipeline.impl.FlinkRemoteBuildPipeline
 import org.apache.streampark.flink.packer.pipeline.impl.FlinkYarnApplicationBuildPipeline;
 
 import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.lang3.StringUtils;
 
 import com.baomidou.mybatisplus.core.conditions.query.LambdaQueryWrapper;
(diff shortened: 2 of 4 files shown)


streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/AppBuildPipeServiceImpl.java
```diff
@@ -234,13 +235,21 @@ public void onStart(PipeSnapshot snapshot) {
...
                     new File(
                         String.format(
                             "%s/%d/%s",
                             Workspace.local().APP_UPLOADS(), app.getTeamId(), app.getJar()));
+                if (!localJar.exists()) {
+                  Resource resource =
+                      resourceService.findByResourceName(app.getTeamId(), app.getJar());
+                  if (resource != null && StringUtils.isNotBlank(resource.getResourcePath())) {
+                    localJar = new File(resource.getResourcePath().split(":")[1]);
+                    uploadJar = appUploads.concat("/").concat(localJar.getName());
```
wolfboys, line 249: We can add the following methods in the Resource entity to conveniently access the name and path.


``` java

    public void setResourcePath(String resourcePath) {
        if (resourcePath == null) {
            throw new IllegalArgumentException("resource path cannot be null.");
        }
        String[] namePath = resourcePath.split(":");
        if (namePath.length != 2) {
            throw new IllegalArgumentException("resource path invalid, format: $name:$path");
        }
        this.resourcePath = resourcePath;
    }

    public String getFileName() {
        if (this.resourcePath == null) {
            return null;
        }
        return resourcePath.split(":")[0];
    }

    public String getFilePath() {
        if (this.resourcePath == null) {
            return null;
        }
        return resourcePath.split(":")[1];
    }

```

ChengJie1053: Ok, thanks for reviewing the code

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/entity/Resource.java
```diff
@@ -82,4 +82,29 @@ public class Resource implements Serializable {
   private transient String sortOrder;
 
   private transient String connector;
+
+  public void setResourcePath(String resourcePath) {
+    if (resourcePath == null) {
+      throw new IllegalArgumentException("resource path cannot be null.");
+    }
+    String[] namePath = resourcePath.split(":");
+    if (namePath.length != 2) {
```
RocMarshal, line 91: @ChengJie1053 Thx for the contribution~
Could we use Utils.require to replace lines `87`, `91` ?

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/ResourceServiceImpl.java
```diff
@@ -204,13 +205,18 @@ public void deleteResource(Resource resource) {
...
-                findResource.getTeamId(),
-                findResource.getResourceName()));
+    String filePath =
+        String.format(
+            "%s/%d/%s",
+            Workspace.local().APP_UPLOADS(),
+            findResource.getTeamId(),
+            findResource.getResourceName());
+
+    if (!new File(filePath).exists() && StringUtils.isNotBlank(findResource.getFilePath())) {
```
wolfboys, line 215: We better use FilesUtils.exists(path...)

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/entity/Resource.java
ChengJie1053: Ok, thanks for reviewing the code

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/ResourceServiceImpl.java
ChengJie1053: Ok, thanks for reviewing the code

wolfboys (approved):

PR #2958, opened 2023-08-21
[type:fix] Refactor AlertType class for fix cannot trigger alert

GOODBOY008:
Fix bug: AlertType's `cachemap` is lack of threading safety. Which can case would not trigger alert action.
@wolfboys @MonsterChenzhuo PTAL.


wolfboys (approved):
LGTM

PR #2965, opened 2023-08-23
[Refactor] Refactor multiple enum classes and update related methods to be more readable

GOODBOY008:
[Refactor] Refactor multiple enum classes and update related methods to be more readable


wolfboys (approved):
LGTM

PR #2970, opened 2023-08-24
Supports specifying  config file when creating  KubernetesClient

ChengJie1053:
Supports specifying  config file when creating  KubernetesClient

### DIFF
--- a/streampark-flink/streampark-flink-kubernetes-v2/streampark-flink-kubernetes-core/src/main/scala/org/apache/streampark/flink/kubernetes/v2/FlinkK8sConfig.scala
+++ b/streampark-flink/streampark-flink-kubernetes-v2/streampark-flink-kubernetes-core/src/main/scala/org/apache/streampark/flink/kubernetes/v2/FlinkK8sConfig.scala
@@ -83,4 +83,12 @@ object FlinkK8sConfig {
     description = "Whether to log the yaml of the generated flink custom resource when submit flink job & cluster"
   )
 
+  // ----- k8s config -----
+
+  val K8S_CONFIG_FILE_PATH: InternalOption = InternalOption(
+    key = "streampark.flink-k8s.config-file-path",
+    defaultValue = "~/.kube/config",
+    classType = classOf[String],
+    description = "K8s Configuration file path"
+  )
 }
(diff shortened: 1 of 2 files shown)


ChengJie1053:
@Al-assad

wolfboys:
Thanks very much for your contribution. Before submitting a pull request, here should be have issue description or email discussion. Please explain the background of the issue clearly, so that the reviewer reviewing the pull request has enough information to better progress the work.

ChengJie1053:
> Thanks very much for your contribution. Before submitting a pull request, here should be have issue description or email discussion. Please explain the background of the issue clearly, so that the reviewer reviewing the pull request has enough information to better progress the work.

Ok, thank you

ChengJie1053:
Currently StreamPark directly uses the system '~ /.kube/config' as the connection credentialfor the Kubernetes cluster. This pr allows users to use custom paths, such as: ’/Users/user/streampark/k8s_config/config’

caicancai:
@Al-assad @wolfboys @ChengJie1053  I think it would be better to support the default path of the system, so that when the error is reported, you can all rely on the upstream. Although I have tried to change it like this, it seems that there is no problem, but it does not guarantee that no error will be reported. There is another situation, that is, a streampark is used by multiple users. If each user creates a .kube/config, I think it is not easy to manage.Of course, this is my point of view, the specifics still have to listen to the boss.

ChengJie1053:
> @Al-assad @wolfboys @ChengJie1053 I think it would be better to support the default path of the system, so that when the error is reported, you can all rely on the upstream. Although I have tried to change it like this, it seems that there is no problem, but it does not guarantee that no error will be reported. There is another situation, that is, a streampark is used by multiple users. If each user creates a .kube/config, I think it is not easy to manage.Of course, this is my point of view, the specifics still have to listen to the boss.

Ok, thanks for reviewing the code. The default value of this configuration is ‘~/.kube/config’, which is the same as the k8s default configuration, and may be required in some cases, such as submitting tasks to multiple k8s clusters

Al-assad:
Hi @ChengJie1053, Thanks for your contribution, but unfortunately I will reject this PR 😣 for the following reasons:
1. Keep kube config reading from the default path, because it is a natural behavior if users run StreamPark on the VM of the kubernetes cluster.
   If StreamPark runs in a K8s POD, then the corresponding kube config is generated by RBAC, and the inserted path for generation is ~/.kube/config in the POD.
2. If it want to provide support for multiple Kubernetes clusters, these alone are not enough.

ChengJie1053:
> Hi @ChengJie1053, Thanks for your contribution, but unfortunately I will reject this PR 😣 for the following reasons:
> 
> 1. Keep kube config reading from the default path, because it is a natural behavior if users run StreamPark on the VM of the kubernetes cluster.
>    If StreamPark runs in a K8s POD, then the corresponding kube config is generated by RBAC, and the inserted path for generation is ~/.kube/config in the POD.
> 2. If it want to provide support for multiple Kubernetes clusters, these alone are not enough.

Ok, thank you for reviewing the code, there are many details that have not been considered

PR #3006, opened 2023-08-31
[ISSUE-2940][Improve] enhance login password protection

### LINKED ISSUE #2940: [Bug] Don't savedown salted password for LDAP login
### Search before asking

### Java Version

1.8+

### Scala Version

2.11.x

### StreamPark Version

latest

### Flink Version

N/A

### deploy mode

None

### What happened

In current LDAP login, streampark will encrypt the password of user's LDAP account, and save down salted password into the database.
It's not a good practice to save it as it might not meet the security or audit requirements.

### Error Exception

_No response_

### Screenshots

_No response_

- [X] Yes I am willing to submit a PR!(您是否要贡献这个PR?)

### Code of Conduct

saLeox:

Issue Number: close #2940
- Don't save the salted password in LDAP authentication;
- Don't allow to update login type when update user;
- Don't allow to reset password for user who doesn't sign in with password.

## Verifying this change

This change is a code cleanup without any test coverage.

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/security/impl/AuthenticatorImpl.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/security/impl/AuthenticatorImpl.java
@@ -59,9 +59,11 @@ public class AuthenticatorImpl implements Authenticator {
 
   private User passwordAuthenticate(String username, String password) {
     User user = usersService.findByName(username);
-    if (user == null || user.getLoginType() != LoginType.PASSWORD) {
-      throw new ApiAlertException(
-          String.format("user [%s] does not exist or can not login with PASSWORD", username));
+    if (user == null) {
+      throw new ApiAlertException(String.format("user [%s] does not exist", username));
+    }
+    if (user.getLoginType() != LoginType.PASSWORD) {
+      throw new ApiAlertException(String.format("user [%s] can not login with PASSWORD", username));
     }
     String salt = user.getSalt();
     password = ShaHashUtils.encrypt(salt, password);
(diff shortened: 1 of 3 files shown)


streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/security/impl/AuthenticatorImpl.java
```diff
@@ -59,9 +59,11 @@ public User authenticate(String username, String password, String loginType) thr
 
   private User passwordAuthenticate(String username, String password) {
     User user = usersService.findByName(username);
-    if (user == null || user.getLoginType() != LoginType.PASSWORD) {
-      throw new ApiAlertException(
-          String.format("user [%s] does not exist or can not login with PASSWORD", username));
+    if (user == null) {
```
wolfboys, line 62: This will cause a bug. In LDAP login mode, if the username doesn't exist when first login, a new user needs to be created. so. we can't simply throw an exception.

saLeox: The `passwordAuthenticate` is used to authenticate the password login, where have to throw error when user doesn't exist, while `ldapAuthenticate` is used for ldap login, and we already have that logic to handle the new user add scenario. Thanks.

wolfboys (approved):
LGTM

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/security/impl/AuthenticatorImpl.java
wolfboys: > The `passwordAuthenticate` is used to authenticate the password login, where have to throw error when user doesn't exist, while `ldapAuthenticate` is used for ldap login, and we already have that logic to handle the new user add scenario. Thanks.

Thank you for your clarification. looks good to me

PR #3016, opened 2023-09-03
[FIX] fix pyflink bug

ChengJie1053:
Fixed bug that pyflink submitted yarn application mode and remote mode tasks


wolfboys (approved):
LGTM

PR #3017, opened 2023-09-04
[Fix] Slim the pac4j package

saLeox:
## Brief change log

- Slim the Pac4j pacakge broung into the project because of SSO integration, and remove the CAS and SAML integration, which is not supported in project;
- Update the LICENSE page for the dependencies involved.


wolfboys (approved):
LGTM

PR #3022, opened 2023-09-05
Optimize the validation logic of flinkDeploymentCR at PR#2966.

Al-assad:

Optimize the validation logic of flinkDeploymentCR at PR#2966 including:
- Use structured Exception instead of RuntimeException.
- Use ZIO side effect instead of throwing exception to interrupt the process.


Al-assad:
cc @ChengJie1053, @wolfboys

PR #3028, opened 2023-09-06
[issues-3027][fix]remove redundant tasks submit logic

alexwxy:
This PR is for fixed bug in issues-3027.
we never submit flink task twice in once action.


wolfboys (approved):
LGTM, Thanks  for your contribution, looking forward to your next contribution  💪

PR #3042, opened 2023-09-11
[Feature] Flink yarn application support  pyflink maven dependency

ChengJie1053:
Importing dependency Jars via maven dependency is supported when submitting pyflink tasks using the yarn application pattern


wolfboys (approved):
LGTM

PR #3145, opened 2023-09-20
[Feature] Sql parser support jar statements and analyze statements

ChengJie1053:
flink-1.15 support JAR Statements
https://nightlies.apache.org/flink/flink-docs-release-1.15/docs/dev/table/sql/jar/

flink-1.16 support ANALYZE Statements
https://nightlies.apache.org/flink/flink-docs-release-1.16/docs/dev/table/sql/analyze/

Sql parser support jar statements and analyze statements


wolfboys:
Thanks for your contribution. JAR statements only work in the [SQL CLI](https://nightlies.apache.org/flink/flink-docs-release-1.15/docs/dev/table/sqlclient/).  StreamPark is different from the implementation of CLI. It is non-command-line interactive, so there are many syntaxes only  work in the [SQL CLI], that StreamPark does not support.

PR #3154, opened 2023-09-20
[Fix] Interceptors add /resource/upload

ChengJie1053:
Web Interceptors add "/resource/upload", and validate python file

### DIFF
--- a/streampark-common/src/main/scala/org/apache/streampark/common/util/FileUtils.scala
+++ b/streampark-common/src/main/scala/org/apache/streampark/common/util/FileUtils.scala
@@ -18,6 +18,8 @@ package org.apache.streampark.common.util
 
 import org.apache.streampark.common.util.ImplicitsUtils._
 
+import org.apache.commons.lang3.StringUtils
+
 import java.io._
 import java.net.URL
 import java.nio.ByteBuffer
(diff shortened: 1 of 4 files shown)


streampark-common/src/main/scala/org/apache/streampark/common/util/FileUtils.scala
```diff
@@ -58,6 +60,9 @@ object FileUtils {
       }) == "504B0304"
   }
 
+  def isPythonFileType(contentType: String): Boolean =
```
wolfboys: "text/x-python" == contentType || "text/x-python-script" == contentType

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/interceptor/UploadFileTypeInterceptor.java
```diff
@@ -44,9 +44,12 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
         MultipartFile multipartFile = multipartRequest.getFile(file);
         ApiAlertException.throwIfNull(
             multipartFile, "File to upload can't be null. Upload file failed.");
-        boolean fileType = FileUtils.isJarFileType(multipartFile.getInputStream());
+        boolean fileType =
```
wolfboys: fileType rename to isJarOrPyFile

streampark-common/src/main/scala/org/apache/streampark/common/util/FileUtils.scala
```diff
@@ -58,6 +64,27 @@ object FileUtils {
       }) == "504B0304"
   }
 
+  def isPythonFileType(contentType: String, input: InputStream): Boolean = {
```
wolfboys, line 67: def isPythonFileType(contentType: String, input: InputStream): Boolean = {
    if (StringUtils.isBlank(contentType) || input == null) {
      throw new RuntimeException("The contentType or inputStream can not be null")
    }

    contentType.contains("text/x-python") && getMimeType(input).equals(MediaType.TEXT_PLAIN.toString)
  }

wolfboys (approved):

PR #3157, opened 2023-09-21
[Fix-3153] Alert information cannot be sent in cancelling state

### LINKED ISSUE #3153: [Bug] Alert information cannot be sent in cancelling state
### Search before asking

### Java Version

1.8

### Scala Version

2.12.x

### StreamPark Version

2.1.1

### Flink Version

1.17.1

### deploy mode

yarn-perjob

### What happened

Alert information cannot be sent in cancelling state.

### Error Exception

_No response_

### Screenshots

_No response_

- [X] Yes I am willing to submit a PR!(您是否要贡献这个PR?)

### Code of Conduct

SbloodyS:
fix #3153

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/watcher/FlinkAppHttpWatcher.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/watcher/FlinkAppHttpWatcher.java
@@ -307,12 +307,21 @@ public class FlinkAppHttpWatcher {
 
     if (flag != null) {
       log.info("FlinkAppHttpWatcher previous state: canceling.");
+      YarnAppInfo yarnAppInfo = httpYarnAppInfo(application);
+      String state = yarnAppInfo.getApp().getFinalStatus();
+      FlinkAppStateEnum flinkAppStateEnum = FlinkAppStateEnum.of(state);
+
+      if (FlinkAppStateEnum.CANCELLING == flinkAppStateEnum) {
+          return;
+      }
+
       if (stopFromEnum.isNone()) {
         log.error(
             "FlinkAppHttpWatcher query previous state was canceling and stopFrom NotFound,savePoint expired!");
         savePointService.expire(application.getId());
+        doAlert(application, flinkAppStateEnum);
       }
-      application.setState(FlinkAppStateEnum.CANCELED.getValue());
+      application.setState(flinkAppStateEnum.getValue());
       cleanSavepoint(application);
       cleanOptioning(optionStateEnum, application.getId());
       doPersistMetrics(application, true);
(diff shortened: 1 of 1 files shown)


streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkRESTAPIWatcher.java
```diff
@@ -505,6 +505,8 @@ and the status is not obtained this time (flink rest server is closed),
         log.error(
             "FlinkRESTAPIWatcher query previous state was canceling and stopFrom NotFound,savePoint expired!");
         savePointService.expire(application.getId());
+        FlinkAppState flinkAppState = FlinkAppState.of(application.getState());
```
wolfboys: Hello, thanks for your contribution. We have checked and found that the job status of the assignment is CANCELED, and it was canceled through StreamaPark. We consider CANCELED to be a normal stoppage status (manually triggered by the user). Our current policy is not to issue any alerts for statuses such as CANCELED and SUCCESS. Only when the assignment FAILED，LOST，CANCELED (not stopped through Streamapark) will an alert be triggered.

lordk911:
When running `yarn application -kill`, the 'Run Status' also displays as 'CANCELED'. If the application is mistakenly deleted, it will cause trouble. So if a Flink application is lost in the resource queue, should we give users an option to issue an alert? Is that better?

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkRESTAPIWatcher.java
SbloodyS: Sorry for late reply. I've been quite busy these days. Will resubmit the code in the next few days...

wolfboys:
> When running `yarn application -kill`, the 'Run Status' also displays as 'CANCELED'. If the application is mistakenly deleted, it will cause trouble. So if a Flink application is lost in the resource queue, should we give users an option to issue an alert? Is that better?

So, flink app on yarn mode, when the flink app status is "cancel"(the app status from Flink metrics), we need to further retrieve the app status from YARN. If the status in YARN is "kill", then we will send an alert.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkRESTAPIWatcher.java
wolfboys: > Sorry for late reply. I've been quite busy these days. Will resubmit the code in the next few days...

I look forward to your driving

lordk911:
> So, flink app on yarn mode, when the flink app status is "cancel"(the app status from Flink metrics), we need to further retrieve the app status from YARN.

thanks, retrieve the app status from YARN again will work out.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkRESTAPIWatcher.java
SbloodyS: Done. PTAL @wolfboys

wolfboys (approved):
LGTM

PR #3164, opened 2023-09-22
[Fix] Fix the issue of Scala calling Mybatis LambdaWrapper with Java function reference

Al-assad:

Fix the issue of Scala calling Mybatis LambdaWrapper with java function reference.

## Brief change log

The column parameter in Mybatis-Plus's LambdaWrapper method likes `eq`, `in`, `set` are of type Java Function and does not actually support any other values except for Java function references.

## Verifying this change


wolfboys (approved):
LGTM

PR #3187, opened 2023-09-25
[Improve] Authentication code optimization

ChengJie1053:
1. Extract common constant
2. Delete org.apache.streampark.console.system.entity.SysLog
3. Add the primary key of the t_team table, which is consistent with mysql

image


wolfboys (approved):
LGTM

PR #3202, opened 2023-09-27
[Improve] Console core module code optimization

ChengJie1053:
[Improve] Console core module code optimization

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/bean/AppBuildDockerResolvedDetail.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/bean/AppBuildDockerResolvedDetail.java
@@ -30,7 +30,6 @@ import javax.annotation.Nullable;
 import java.util.Comparator;
 import java.util.Date;
 import java.util.List;
-import java.util.Objects;
 import java.util.stream.Collectors;
 
 /** View for DockerResolveProgress */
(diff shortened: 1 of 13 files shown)


streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/bean/AppBuildDockerResolvedDetail.java
```diff
@@ -43,7 +44,7 @@ public class AppBuildDockerResolvedDetail {
 
   public static AppBuildDockerResolvedDetail of(@Nullable DockerResolvedSnapshot snapshot) {
     AppBuildDockerResolvedDetail detail = new AppBuildDockerResolvedDetail();
-    if (snapshot == null) {
+    if (Objects.isNull(snapshot)) {
```
wolfboys: If there is a lambda expression in this code, it will be more concise to use Objects.isNull, e.g:
```
List<String> list = Arrays.asList("a", "b", null, null);
        List<String> list2 = list.stream()
                .filter(Objects::nonNull)
                .collect(Collectors.toList())
```
.Conversely, it is not recommended to use Objects.isNull; it is better to directly check if the object is null, such as if (obj == null).

ChengJie1053: Ok, thanks for reviewing the code

wolfboys (approved):
LGTM

PR #3237, opened 2023-10-13
[Fix-3226] If the user does not have any more attributes, they will not be able to log in.

### LINKED ISSUE #3226: [Bug] [LDAP]If the user does not have any more attributes, they will not be able to log in.
### Search before asking

### Java Version

1.8

### Scala Version

2.11.x

### StreamPark Version

2.1.1

### Flink Version

1.17

### deploy mode

None

### What happened

I use openldap, by default there are no additional attributes such as mail. In this case, even if the password is correct, you cannot log in unless an additional attribute is added to the corresponding user. When using LDAP for authentication, can we only verify the password?

https://github.com/apache/incubator-streampark/blob/095be65b7ac9ee7c71c4813fbbf0b2fc7b6be422/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/security/impl/LdapService.java#L100

The code runs to this point. If there are no additional attributes, the password will not be verified. However, if there are multiple attributes, does it require multiple password verifications?

https://github.com/apache/incubator-streampark/blob/095be65b7ac9ee7c71c4813fbbf0b2fc7b6be422/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/security/impl/AuthenticatorImpl.java#L80

Email may not be necessary.

### Error Exception

_No response_

### Screenshots

_No response_

- [ ] Yes I am willing to submit a PR!(您是否要贡献这个PR?)

### Code of Conduct

SbloodyS:
fix #3226 

Remove email check logic in ldap auth since email is not mandatory when creating a user.

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/security/impl/LdapService.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/security/impl/LdapService.java
@@ -70,7 +70,7 @@ public class LdapService {
    *
    * @param userId user identity id
    * @param userPwd user login password
-   * @return user email
+   * @return boolean ldapLoginStatus
    */
   public boolean ldapLogin(String userId, String userPwd) {
 
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/security/impl/LdapService.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/security/impl/LdapService.java
@@ -72,7 +72,7 @@ public class LdapService {
    * @param userPwd user login password
    * @return user email
    */
-  public String ldapLogin(String userId, String userPwd) {
+  public boolean ldapLogin(String userId, String userPwd) {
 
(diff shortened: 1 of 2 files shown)


streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/security/impl/LdapService.java
```diff
@@ -104,18 +104,18 @@ public String ldapLogin(String userId, String userPwd) {
             new InitialDirContext(ldapEnv);
           } catch (Exception e) {
             log.warn("invalid ldap credentials or ldap search error", e);
-            return null;
+            return false;
           }
           Attribute attr = attrs.next();
-          if (attr.getID().equals(ldapEmailAttribute)) {
-            return (String) attr.get();
+          if (attr.getID().equals(ldapUserIdentifyingAttribute)) {
```
wolfboys, line 110: the code can be optimized  `return attr.getID().equals(ldapUserIdentifyingAttribute)`

wolfboys (approved):
LGTM

PR #3258, opened 2023-10-21
[Improve] Improve streampark-common module base

VampireAchao:
see:  https://rules.sonarsource.com/java/RSPEC-3457/

> Because printf-style format strings are interpreted at runtime, rather than validated by the compiler, they can contain errors that result in the wrong strings being created. This rule statically validates the correlation of printf-style format strings to their arguments when calling the format(...) methods of java.util.Formatter, java.lang.String, java.io.PrintStream, MessageFormat, and java.io.PrintWriter classes and the printf(...) methods of java.io.PrintStream or java.io.PrintWriter classes.

Issue Number: close #xxx 

## Verifying this change

*(or)*

*(or)*


wolfboys (approved):
LGTM

PR #3274, opened 2023-10-23
[Improve][All] Improvement of wrapper data type

wuchunfu:
All POJO class properties must use wrapper data types

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/bean/EmailConfig.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/bean/EmailConfig.java
@@ -38,7 +38,7 @@ public class EmailConfig {
   private String from;
   private String userName;
   private String password;
-  private boolean ssl;
+  private Boolean ssl;
 
   /**
    * Constructs the EmailConfig object from the given settings map.
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/bean/ResponseResult.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/bean/ResponseResult.java
@@ -21,7 +21,7 @@ import lombok.Data;
 
 @Data
 public class ResponseResult<T> {
-  private int status;
+  private Integer status;
   private String msg;
   private T result;
 }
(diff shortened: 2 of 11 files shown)


wolfboys:
bugs will occur, int default value is 0, Integer is null. boolean default value  is false, Boolean default value  null ...

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/bean/EmailConfig.java
```diff
@@ -38,7 +38,7 @@ public class EmailConfig {
   private String from;
   private String userName;
   private String password;
-  private boolean ssl;
+  private Boolean ssl;
```
caicancai, line 41: Why choose Boolean instead of boolean. I think boolean is better

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/bean/ResponseResult.java
```diff
@@ -21,7 +21,7 @@
 
 @Data
 public class ResponseResult<T> {
-  private int status;
+  private Integer status;
```
caicancai, line 24: Why choose Interger over Int. I think Int is better

caicancai:
> bugs will occur, int default value is 0, Integer is null. boolean default value is false, Boolean default value null ...

+1

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/bean/EmailConfig.java
wuchunfu: @caicancai Thank you for your review, I agree with your point of view and I will close this pr

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/bean/ResponseResult.java
wuchunfu: @caicancai Thank you for your review, I agree with your point of view and I will close this pr

wuchunfu:
> bugs will occur, int default value is 0, Integer is null. boolean default value is false, Boolean default value null ...

@wolfboys Thank you for your review, I agree with your point of view and I will close this pr

PR #3276, opened 2023-10-24
[Fix] sqlDialect requires capitalization 

ChengJie1053:
'default' needs to be uppercase, otherwise it won't match


streampark-common/src/main/java/org/apache/streampark/common/Constant.java
```diff
@@ -24,6 +24,8 @@ private Constant() {}
 
   public static final String DEFAULT = "default";
 
+  public static final String DEFAULT_UPPER_CASE = "DEFAULT";
```
wolfboys, line 27: I don't think we need this constant. 😕

PR #3277, opened 2023-10-24
[Fix] SqlDialect `DEFAULT` requires capitalization

ChengJie1053:
[Fix] SqlDialect `DEFAULT` requires capitalization


wolfboys (approved):
LGTM

PR #3289, opened 2023-10-25
[Improve] Rename 'streampark.yarn.http-auth' configuration option from 'sample' to 'simple'

ylw-github:
This PR is intended to rename the 'streampark.yarn.http-auth' configuration option from 'sample' to 'simple'. The change is made to ensure compatibility with Hadoop, as Hadoop does not support the 'sample' authentication mode. This modification prevents potential issues and confusion in the future.


wolfboys (approved):
LGTM

PR #3313, opened 2023-11-05
[Improve] submit flink job userclassPaths improvement

wolfboys:
[Improve] submit flink job userclassPaths improvement


streampark-flink/streampark-flink-client/streampark-flink-client-api/src/main/scala/org/apache/streampark/flink/client/bean/SubmitRequest.scala
```diff
@@ -68,9 +70,15 @@ case class SubmitRequest(
...
+  lazy val classPaths: List[URL] = {
+    val path = s"${Workspace.local.APP_WORKSPACE}/$id/lib"
+    val lib = Try(new File(path).listFiles().map(_.toURI.toURL).toList).getOrElse(List.empty[URL])
+    flinkVersion.flinkLibs ++ lib
+  }
+
   lazy val flinkSQL: String = extraParameter.get(KEY_FLINK_SQL()).toString
 
-  lazy val allowNonRestoredState = Try(
+  lazy val allowNonRestoredState: Boolean = Try(
```
caicancai, line 81: scala has default type inference, and I'm not sure if it needs to be declared as a Boolean type

caicancai:
cc @Al-assad

caicancai (approved):

caicancai:
I'm curious if Lazy Val needs to declare variable types, since Scala has automatic type inference

wolfboys:
> I'm curious if Lazy Val needs to declare variable types, since Scala has automatic type inference

The "lazy" here has a literal meaning. In the context of a class, it refers to a field that is only initialized when it is actually used.

caicancai:
> > I'm curious if Lazy Val needs to declare variable types, since Scala has automatic type inference
> 
> The "lazy" here has a literal meaning. In the context of a class, it refers to a field that is only initialized when it is actually used.

Thanks for the reply, I have one more question, about the type display of Scala, is it uniform in the SP code? Whether to show or not to show the type, in fact, this does not affect any code in the dev, this is just a small suggestion of mine @Al-assad

Al-assad (approved):

Al-assad:
> > > I'm curious if Lazy Val needs to declare variable types, since Scala has automatic type inference
> 
> > 
> 
> > The "lazy" here has a literal meaning. In the context of a class, it refers to a field that is only initialized when it is actually used.
> 
> 
> 
> Thanks for the reply, I have one more question, about the type display of Scala, is it uniform in the SP code? Whether to show or not to show the type, in fact, this does not affect any code in the dev, this is just a small suggestion of mine @Al-assad 

Type declarations in Scala are not mandatory, they are only required if you provide the compiler with a type annotation

caicancai:
> > > > I'm curious if Lazy Val needs to declare variable types, since Scala has automatic type inference
> > 
> > 
> > > 
> > 
> > 
> > > The "lazy" here has a literal meaning. In the context of a class, it refers to a field that is only initialized when it is actually used.
> > 
> > 
> > Thanks for the reply, I have one more question, about the type display of Scala, is it uniform in the SP code? Whether to show or not to show the type, in fact, this does not affect any code in the dev, this is just a small suggestion of mine @Al-assad
> 
> Type declarations in Scala are not mandatory, they are only required if you provide the compiler with a type annotation

I might be trying to unify the way streampark is written, but at the moment it seems irrelevant

caicancai:
LGTM + 1

PR #3315, opened 2023-11-05
[Improve][Flink] Improve deployreponse parameters

caicancai:
Add @Nullable to address because v2 does return an empty parameter at the moment

Issue Number: close #xxx 

## Verifying this change

*(or)*

*(or)*

### DIFF
--- a/streampark-flink/streampark-flink-client/streampark-flink-client-api/src/main/scala/org/apache/streampark/flink/client/bean/DeployResponse.scala
+++ b/streampark-flink/streampark-flink-client/streampark-flink-client-api/src/main/scala/org/apache/streampark/flink/client/bean/DeployResponse.scala
@@ -21,7 +21,4 @@ import javax.annotation.Nullable
 
 import java.util.{Map => JavaMap}
 
-case class DeployResponse(
-    @Nullable address: String = "",
-    clusterId: String,
-    flinkConfig: JavaMap[String, String])
+case class DeployResponse(@Nullable address: String = "", clusterId: String)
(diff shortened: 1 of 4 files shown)


caicancai:
cc @wolfboys

caicancai:
For ② I think it is OK, ① I am not sure if it is appropriate to make such a change

streampark-flink/streampark-flink-client/streampark-flink-client-api/src/main/scala/org/apache/streampark/flink/client/bean/DeployResponse.scala
```diff
@@ -17,4 +17,11 @@
...
 
-case class DeployResponse(address: String, clusterId: String)
+import javax.annotation.Nullable
+
+import java.util.{Map => JavaMap}
+
+case class DeployResponse(
+    @Nullable address: String = "",
+    clusterId: String,
+    flinkConfig: JavaMap[String, String])
```
wolfboys: Can you tell me why we added this field? Where do we use it?

caicancai: I am also very confused about this part, mainly to align submitrequest, I do not know how submitrquest uses this parameter

caicancai: The flinkconfig parameter is very important during the deployment process

wolfboys: > The flinkconfig parameter is very important during the deployment process

I see that we don't currently need this parameter, so I think it would be more appropriate to add it when we actually need it.

caicancai: ok

caicancai: done

caicancai:
cc @wolfboys Add @Nullable to address because v2 does return an empty parameter at the moment

wolfboys (approved):
LGTM

PR #3329, opened 2023-11-10
[Bug] deploy flink job on yarn, get state bug fixed.

wolfboys:
[Bug] deploy flink job on yarn, get state bug fixed.


caicancai:
@Al-assad If you have time, could you review it? Thanks

Al-assad (approved):

PR #3332, opened 2023-11-10
 [Fix] Fix yarn session create jobGraph failed

ChengJie1053:
[Fix] Fix yarn session create jobGraph failed


wolfboys:
fixed #3329  [details](https://github.com/apache/incubator-streampark/blob/97cc84c77695814d311bd495921095d0c1405856/streampark-flink/streampark-flink-client/streampark-flink-client-core/src/main/scala/org/apache/streampark/flink/client/trait/FlinkClientTrait.scala#L246)

PR #3381, opened 2023-12-07
[Improve] improve startup log annotation

caicancai:
Very small improvement


streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/StreamParkConsoleBootstrap.java
```diff
@@ -33,9 +33,11 @@
  *    ___/ / /_/ /  /  __/ /_/ / / / / / / /_/ / /_/ / /  / ,<
  *   /____/\__/_/   \___/\__,_/_/ /_/ /_/ ____/\__,_/_/  /_/|_|
  *                                     /_/
- *
+ *   Version:  2.2.0-SNAPSHOT
```
wolfboys, line 36: It is not recommended to modify this.

wolfboys:
Thanks for your contribution. I do not agree with this PR and do not recommend changes.

PR #3427, opened 2023-12-25
[Improve] Use StringUtils.isNotBlank instead of StringUtils.isNotEmpty

xxzuo:
Use StringUtils.isNotBlank instead of StringUtils.isNotEmpty


wolfboys (approved):
LGTM

PR #3443, opened 2024-01-01
[Feature] apache-rat-plugin support

wolfboys:
### [Feature] apache-rat-plugin support

The apache-rat-plugin can help us better check the license header of project source files.  It is of great assistance to project release and compliance.


caicancai (approved):

caicancai:
Perhaps we should describe in this PR why we changed it so that it would be better for archiving.

GOODBOY008 (approved):
LGTM

PR #3469, opened 2024-01-08
[Improve] flink env check improvement

wolfboys:
[Improve] flink env check improvement


streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/enums/FlinkEnvCheckEnum.java
```diff
@@ -24,16 +24,20 @@
...
 
   /* FLINK_HOME path invalid */
-  INVALID(-1),
+  INVALID_PATH(-1),
 
   /* this add/update operation ok */
   OK(0),
 
-  /* defined flink name repeated */
+  /** name repeated */
```
zhengkezhou1:
```java

 / * name repeated */

zhengkezhou1 (changes requested):

GOODBOY008 (approved):
LGTM

PR #3528, opened 2024-02-01
[Improve] Optimize equals method conflicts

linweijiang:
Improve the conflict issue between the equals method in lombok's `@Data`  annotation and the equals method in the same class during the install packaging process.

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/bean/Dependency.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/bean/Dependency.java
@@ -26,7 +26,7 @@ import org.apache.streampark.flink.packer.maven.Artifact;
 import org.apache.streampark.flink.packer.maven.DependencyInfo;
 
 import com.fasterxml.jackson.core.type.TypeReference;
-import lombok.Data;
+import lombok.Getter;
 import lombok.SneakyThrows;
 
 import java.io.File;
@@ -36,7 +36,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-@Data
+@Getter
 public class Dependency {
   private List<MavenPom> pom = Collections.emptyList();
   private List<String> jar = Collections.emptyList();


caicancai:
Can you tell me why you changed it like this?

linweijiang:
> Can you tell me why you changed it like this?

When I execute the 'mvn install' command in a Windows environment, conflicts arise due to the lombok's `@Data` annotation, which by default implements an equals method that collides with the manually written equals method in the current class, resulting in the following exception: 
`Not generating hashCode: One of equals or hashCode exists. You should either write both of these or none of these (in the latter case, lombok generates them).`

caicancai:
> > Can you tell me why you changed it like this?
> 
> When I execute the 'mvn install' command in a Windows environment, conflicts arise due to the lombok's `@Data` annotation, which by default implements an equals method that collides with the manually written equals method in the current class, resulting in the following exception: `Not generating hashCode: One of equals or hashCode exists. You should either write both of these or none of these (in the latter case, lombok generates them).`

OK, thank you for your explanation and contribution. I will verify it and if there is no problem I will merge it. Can you please add the reason to the PR description?

linweijiang:
> > > Can you tell me why you changed it like this?
> > 
> > 
> > When I execute the 'mvn install' command in a Windows environment, conflicts arise due to the lombok's `@Data` annotation, which by default implements an equals method that collides with the manually written equals method in the current class, resulting in the following exception: `Not generating hashCode: One of equals or hashCode exists. You should either write both of these or none of these (in the latter case, lombok generates them).`
> 
> OK, thank you for your explanation and contribution. I will verify it and if there is no problem I will merge it. Can you please add the reason to the PR description?

I have added the description information. Can you help me to check it? This is my first time submitting a PR to our project, thank you.

VampireAchao (changes requested):
Thank you for your submission. I've reviewed the changes and I'd like to suggest an alternative approach that might align better with our goals of minimizing code intrusion while retaining the flexibility of our model's accessors. Instead of changing @Data to @Getter, consider applying the lombok.experimental.Tolerate annotation on the equals method. This approach is less invasive as it preserves the setters, which might be beneficial for our use cases where setter methods are needed. Moreover, the Tolerate annotation is specifically designed for scenarios like this, allowing us to override method signatures without losing the benefits of Lombok's automatic generation features. I believe this will keep our code both clean and functional. Let me know if you have any questions or need further clarification on implementing this. Thanks!

linweijiang:
> Thank you for your submission. I've reviewed the changes and I'd like to suggest an alternative approach that might align better with our goals of minimizing code intrusion while retaining the flexibility of our model's accessors. Instead of changing @DaTa to @getter, consider applying the lombok.experimental.Tolerate annotation on the equals method. This approach is less invasive as it preserves the setters, which might be beneficial for our use cases where setter methods are needed. Moreover, the Tolerate annotation is specifically designed for scenarios like this, allowing us to override method signatures without losing the benefits of Lombok's automatic generation features. I believe this will keep our code both clean and functional. Let me know if you have any questions or need further clarification on implementing this. Thanks!

Thank you for your advice. After examining the usage of the class, I've noticed that there are no calls to functionalities provided by the `@Data` annotation other than the `@Getter`. Hence, it seems that only incorporating the `@Getter` annotation might be a better choice.

wolfboys (approved):
LGTM

wolfboys:
good job, Thanks for your contribution, Looking forward to your next contribution! 🙌

VampireAchao:
Thank you for your detailed review and consideration. I agree with your conclusion that utilizing only the @Getter annotation is the best approach given the current use of the class. Your analysis makes sense, and I support moving forward with this solution. I'll approve the PR based on this decision. LGTM!

VampireAchao (approved):

PR #3529, opened 2024-02-01
[Fix] The initialization of the configuration comes before the bean is loaded

### LINKED ISSUE #3494: [Bug] Before initializing springboot, get enable-v2 and it will always be the default value
### Search before asking

### Java Version

_No response_

### Scala Version

2.12.x

### StreamPark Version

dev

### Flink Version

flink all version

### deploy mode

kubernetes-session

### What happened

When I wanted to start flink on k8s V2

I changed streampark.flink-k8s.enable-v2 to `true` in the configuration file but found that

`isV2Enabled` is of type `val`. After the class is initialized, it will be assigned false directly, 
image

As a result, the values inside are refreshed even after springboot starts.

image

However, there are two methods to get this value in advance, so in the initialization phase, you cannot get the value of springboot's yaml configuration, you get the initialization value (false)

The following is a method to use in advance when starting up

image
image

### Error Exception

_No response_

### Screenshots

_No response_

- [x] Yes I am willing to submit a PR!(您是否要贡献这个PR?)

### Code of Conduct

Xuxiaotuan:
The initialization of the configuration comes before the bean is loaded

Issue Number: close #3494 

## Brief change log

The initial configuration load consists of two parts:
 - 1. one is the native java and scala default configuration
 - 2. the other is the environment variable passed into the ApplicationContext via yaml or parameter configuration

The current streampark flushes the yaml configuration into the global configuration cache after springboot starts. 

As a result, during bean initialization, if you want to use yaml or parameter configuration to pass the value, 

it will fail, and the default value will be used instead.

I recommend that the initialization of the configuration cache be advanced to before the bean is loaded

## Verifying this change

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/StreamParkConsoleBootstrap.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/StreamParkConsoleBootstrap.java
@@ -25,6 +25,8 @@ import org.springframework.boot.builder.SpringApplicationBuilder;
 import org.springframework.scheduling.annotation.EnableScheduling;
 
 /**
+ *
+ *
  * <pre>
  *
  *      _____ __                                             __
@@ -46,9 +48,9 @@ import org.springframework.scheduling.annotation.EnableScheduling;
 @EnableScheduling
 public class StreamParkConsoleBootstrap {
 
-    public static void main(String[] args) {
-        new SpringApplicationBuilder(StreamParkConsoleBootstrap.class)
-            .initializers(new EnvApplicationContextInitializer())
-            .run(args);
-    }
+  public static void main(String[] args) {
+    new SpringApplicationBuilder(StreamParkConsoleBootstrap.class)
+        .initializers(new EnvApplicationContextInitializer())
+        .run(args);
+  }
 }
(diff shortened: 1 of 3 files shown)


caicancai:
@Al-assad If you have time, can you help me review this PR? Thank you.

caicancai:
@Xuxiaotuan Please fix cli

caicancai:
It seems that there are still some problems. You can wait for the database problem to be fixed before completing your PR. You can first add WIP to your PR title to indicate that it is in progress.
By the way, your commit message needs to be changed to English. Thank you.

wolfboys (approved):
LGTM

caicancai:
@wolfboys why merage？cli fail

wolfboys:
> @wolfboys why merage？cli fail

Thanks for your feedback, the CI failure was not caused by this PR. I will fix it soon. Happy New Year! 🎉

caicancai:
Thank you for your reply, but this PR has a WIP tag. When we merge it, we should remove the WIP tag, and its commit message is in Chinese. I think it is unreasonable. We should pay attention to such issues in the future.
Finally, happy new year🎉

wolfboys:
> Thank you for your reply, but this PR has a WIP tag. When we merge it, we should remove the WIP tag, and its commit message is in Chinese. I think it is unreasonable. We should pay attention to such issues in the future. Finally, happy new year🎉

👌👌👌

PR #3550, opened 2024-02-16
[Improve] Add dependency support to unit tests on streampark-common_2.12 

zhengkezhou1:
Adding Dependency Support to Unit Tests

## Brief change log

*Add mockito-scala.version to root pom.xml*
*Add  unit test dependencies to streampark-common_2.12 pom.xml*
## Verifying this change

### DIFF
--- a/streampark-common/pom.xml
+++ b/streampark-common/pom.xml
@@ -31,6 +31,13 @@
 
         <!-- test -->
         <dependency>
+            <groupId>org.scalatest</groupId>
+            <artifactId>scalatest_${scala.binary.version}</artifactId>
+            <version>${scalatest.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-engine</artifactId>
             <version>${jupiter.version}</version>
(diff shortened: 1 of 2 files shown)


zhengkezhou1:
@caicancai pls take a look

streampark-common/pom.xml
```diff
@@ -30,6 +30,20 @@
...
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-scala_2.12</artifactId>
+            <version>${mockito-scala.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.scalatest</groupId>
+            <artifactId>scalatest_2.12</artifactId>
```
caicancai: it should scalatest_${scala.binary.version}

streampark-common/pom.xml
```diff
@@ -30,6 +30,20 @@
     <dependencies>
 
         <!-- test -->
+        <dependency>
```
caicancai, line 33: why add it？If it is used in the test use case, please go with the test use case, otherwise the developer will not be able to understand it.

caicancai (changes requested):

caicancai (approved):
LGTM

PR #3557, opened 2024-02-16
[Imprpve] JDBC "like" syntax is compatible with mysql and pgsql

wolfboys:
[Imprpve] JDBC "like" syntax is compatible with mysql and pgsql


VampireAchao:
LGTM!

VampireAchao (approved):
LGTM!

PR #3564, opened 2024-02-19
[ISSUES-3534] Fix path resolution exception when submitting tasks in remote mode on Windows environment

linweijiang:
In a Windows environment, replace the `\` in the jar path with `/`.

### DIFF
--- a/streampark-flink/streampark-flink-client/streampark-flink-client-core/src/main/scala/org/apache/streampark/flink/client/trait/FlinkClientTrait.scala
+++ b/streampark-flink/streampark-flink-client/streampark-flink-client-core/src/main/scala/org/apache/streampark/flink/client/trait/FlinkClientTrait.scala
@@ -102,7 +102,11 @@ trait FlinkClientTrait extends Logger {
         }
       case _ =>
         if (submitRequest.userJarFile != null) {
-          val uri = PackagedProgramUtils.resolveURI(submitRequest.userJarFile.getAbsolutePath)
+          var jarPath = submitRequest.userJarFile.getAbsolutePath
+          if (Utils.isWindows) {
+            jarPath = StringUtils.replace(jarPath, "\\", "/")
+          }
+          val uri = PackagedProgramUtils.resolveURI(jarPath)
           val programOptions = ProgramOptions.create(commandLine)
           val executionParameters = ExecutionConfigAccessor.fromProgramOptions(
             programOptions,


caicancai:
Whether windows can use wsl to avoid this problem

streampark-flink/streampark-flink-client/streampark-flink-client-core/src/main/scala/org/apache/streampark/flink/client/trait/FlinkClientTrait.scala
```diff
@@ -102,7 +102,11 @@ trait FlinkClientTrait extends Logger {
         }
       case _ =>
         if (submitRequest.userJarFile != null) {
-          val uri = PackagedProgramUtils.resolveURI(submitRequest.userJarFile.getAbsolutePath)
+          var jarPath = submitRequest.userJarFile.getAbsolutePath
```
wolfboys, line 105: Thanks very much for your contribution. It is not recommended to make changes to the code here. You can make changes to the path in the ApplicationService as follows: 

before:
image

after:
image

You can do a global search to improve all the path concat in this way. Looking forward to your updates! 🌟

fuchanghai:
i agree with @linweijiang ,because  you cant know how much ``` \\ ```  it has ,and you are not sure about the hierarchical structure of the address of the file where the jar package is located. so  global replacement is a better way. but we'd better write this replacement method in the util class. cc @caicancai @wolfboys  WDYT？

SbloodyS:
Currently, most big data projects are not compatible with windows, including Spark and Flink. To be compatible with Windows, it may require a lot of extra work for maintainers. So I think this job is unnecessary.

fuchanghai (commented):
+0

linweijiang:
> Currently, most big data projects are not compatible with windows, including Spark and Flink. To be compatible with Windows, it may require a lot of extra work for maintainers. So I think this job is unnecessary.

Thanks for sharing your valuable ideas. But I believe that whether it's Flink or Spark, users should be able to run them normally in a Windows environment and conduct local debugging. I think that enabling users to perform local debugging in a Windows environment is an extremely valuable feature.

caicancai:
My idea is if you can solve this problem from wsl, I tried using wsl debug spark with no problem.
It's not a big change, I can accept it.

wolfboys:
I think we just need to solve the compatibility issue with paths on Windows, This will be very convenient for developers on Windows.

streampark-flink/streampark-flink-client/streampark-flink-client-core/src/main/scala/org/apache/streampark/flink/client/trait/FlinkClientTrait.scala
linweijiang: > Thanks very much for your contribution. It is not recommended to make changes to the code here. You can make changes to the path in the ApplicationService as follows:
> 
> before: image
> 
> after: image
> 
> You can do a global search to improve all the path concat in this way. Looking forward to your updates! 🌟

hi~ @wolfboys , tks for you advice. Can you successfully test it locally? I attempted to debug it in this manner locally but failed, and it's still showing the same error as before.

caicancai:
closing

PR #3568, opened 2024-02-20
[BUG][FLINK-DEPEND]Fix flink lib to exclude dependencies that do not end in.jar 

zhilinli123:

Issue Number: close 
### What issues are fixed

image
image

This problem was resolved when I eliminated the need to rely on.jar endings. Many developers get used to the name.bak for easy backup, but applications shouldn't load it 

## Verifying this change

*(or)*

*(or)*


wolfboys (approved):
LGTM, good job. ☺️ We are looking forward to your next contribution 🌟

PR #3572, opened 2024-02-26
[Improve] mysql and postgre ddl current_timestamp syntax compatibility improvement.

### LINKED ISSUE #3570: [Bug] mysql ddl "on update current_timestamp" compatibility improvements
### Search before asking

### Java Version

all

### Scala Version

2.11.x

### StreamPark Version

all

### Flink Version

all

### deploy mode

None

### What happened

Currently, there is a syntax "ON UPDATE CURRENT_TIMESTAMP" in StreamPark's MySQL DDL, This syntax is only supported starting from MySQL version 5.6.5. For users whose MySQL version is lower than 5.6, it will cause the execution ddl failed.

### Error Exception

_No response_

### Screenshots

_No response_

- [ ] Yes I am willing to submit a PR!(您是否要贡献这个PR?)

### Code of Conduct

joyCurry30:

Issue Number: close #3570  

Use @TableField to replace the SQL syntax ON UPDATE CURRENT_TIMESTAMP.

## Verifying this change

*(or)*

*(or)*

 - Dependencies (does it add or upgrade a dependency): (yes / no)  no

### DIFF
--- a/streampark-console/streampark-console-service/src/main/assembly/script/schema/mysql-schema.sql
+++ b/streampark-console/streampark-console-service/src/main/assembly/script/schema/mysql-schema.sql
@@ -33,7 +33,7 @@ create table `t_app_backup` (
   `version` int default null,
   `path` varchar(128) collate utf8mb4_general_ci default null,
   `description` varchar(255) collate utf8mb4_general_ci default null,
-  `create_time` datetime not null default current_timestamp comment 'create time',
+  `create_time` datetime default null comment 'create time',
   primary key (`id`) using btree
 ) engine=innodb auto_increment=100000 default charset=utf8mb4 collate=utf8mb4_general_ci;
 
@@ -86,8 +86,8 @@ create table `t_flink_app` (
   `available_slot` int default null,
   `option_state` tinyint default null,
   `tracking` tinyint default null,
-  `create_time` datetime not null default current_timestamp comment 'create time',
-  `modify_time` datetime not null default current_timestamp comment 'modify time',
+  `create_time` datetime default null comment 'create time',
+  `modify_time` datetime default null comment 'modify time',
   `option_time` datetime default null,
   `release` tinyint default 1,
   `build` tinyint default 1,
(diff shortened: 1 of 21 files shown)


streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/entity/User.java
```diff
@@ -70,8 +72,10 @@ public class User implements Serializable {
   @NotBlank(message = "{required}")
   private String status;
 
+  @TableField(fill = FieldFill.INSERT)
   private Date createTime;
 
+  @TableField(update = "now()", fill = FieldFill.INSERT_UPDATE)
```
wolfboys, line 78: It is not advisable to use annotations. This will trigger an update whenever there is a change in the records in the table, which does not meet our expectations. The `modifyTime` is only updated when users manually modify entity info on the frontend. Often, StremaPark will modify records in the table, e.g: When the system detects a change in the  job  status. that needs to be synchronized to the database. In these cases, we do not want to update the `modifyTime`.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/system/entity/User.java
```diff
@@ -70,8 +72,10 @@ public class User implements Serializable {
   @NotBlank(message = "{required}")
   private String status;
 
+  @TableField(fill = FieldFill.INSERT)
```
wolfboys: It is suggested to manually set the value when new entity to be save.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/entity/Application.java
```diff
@@ -202,10 +203,12 @@ public class Application implements Serializable {
 
   private String description;
 
+  @TableField(fill = FieldFill.INSERT)
   private Date createTime;
 
   private Date optionTime;
 
+  @TableField(update = "now()", fill = FieldFill.INSERT_UPDATE)
   private Date modifyTime;
```
wolfboys: It is not advisable to use annotations. This will trigger an update whenever there is a change in the records in the table, which does not meet our expectations. The `modifyTime` is only updated when users manually modify entity info on the frontend. Often, StremaPark will modify records in the table, e.g: When the system detects a change in the job status. that needs to be synchronized to the database. In these cases, we do not want to update the `modifyTime`.

wolfboys (approved):
LGTM

Issue #3583, opened 2024-02-27
[Improve] docker|email settings improvements


wolfboys:
### Search before asking

- [X] I had searched in the [feature](https://github.com/apache/incubator-streampark/issues?q=is%3Aissue+label%3A%22Feature%22) and found no similar feature requirement.

### Description

Currently in StreamPark, on the Docker info settings page, each docker setting item is displayed on the page, each items can be edited and there is no verification. so, we need improvement this: The Docker settings should only have one entry point. When clicked for editing, there will be a form with fields for each setting. Validation of this information is required when the form is submitted.  email settings also need improvement。

<img width="1389" alt="image" src="https://github.com/apache/incubator-streampark/assets/13284744/59b5498d-5c75-4f11-81c2-cbeb4b0b6b82">

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wolfboys:
done

PR #3590, opened 2024-02-28
[Improve] Check email Setting before update

zhengkezhou1:

Link to :  #3583 
*Before performing an update to the configuration of the email alert message, check that the configuration is correct Simply send a test email*

## Verifying this change

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/SettingServiceImpl.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/SettingServiceImpl.java
@@ -17,7 +17,6 @@
 
 package org.apache.streampark.console.core.service.impl;
 
-import org.apache.streampark.console.base.exception.AlertException;
 import org.apache.streampark.console.core.bean.DockerConfig;
 import org.apache.streampark.console.core.bean.MavenConfig;
 import org.apache.streampark.console.core.bean.ResponseResult;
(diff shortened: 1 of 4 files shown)


zhengkezhou1:
Hi @wolfboys, sry for the late commit,  pls take a look.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/SettingServiceImpl.java
```diff
@@ -148,8 +153,24 @@ public ResponseResult checkDocker(DockerConfig dockerConfig) {
   }
 
   @Override
-  public ResponseResult checkEmail(SenderEmail senderEmail) {
-    // TODO check
+  public ResponseResult checkEmail(SenderEmail senderEmail) throws MessagingException {
```
wolfboys: Do not throw exceptions to the Controller layer, return the validation result instead.

```
 @Override
    public ResponseResult checkEmail(SenderEmail senderEmail) {
        ResponseResult result = new ResponseResult();
        Properties props = new Properties();
        props.put("mail.smtp.auth", "true");
        props.put("mail.smtp.starttls.enable", "true");
        props.put("mail.smtp.host", senderEmail.getHost());
        props.put("mail.smtp.port", senderEmail.getPort());

        Session session = Session.getInstance(props);
        try {
            Transport transport = session.getTransport("smtp");
            transport.connect(
                senderEmail.getHost(), senderEmail.getUserName(), senderEmail.getPassword());
            transport.close();
            result.setStatus(200);
        } catch (MessagingException e) {
            result.setStatus(500);
            result.setMsg("connect to target mail server failed: " + e.getMessage());
        }
        return result;
    }
```

PR #3592, opened 2024-02-29
[Improve] introduce a docker config check method

zhengkezhou1:

Link to :  #3583 
**Before making updates to the docker configuration, we introduce a check to test that incoming parameters can be concatenated with the registered address**

## Verifying this change

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/SettingServiceImpl.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/SettingServiceImpl.java
@@ -26,6 +26,12 @@ import org.apache.streampark.console.core.mapper.SettingMapper;
 import org.apache.streampark.console.core.service.SettingService;
 
 import org.apache.commons.lang3.StringUtils;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpPost;
+import org.apache.http.entity.ContentType;
+import org.apache.http.entity.StringEntity;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClients;
 
 import com.baomidou.mybatisplus.core.conditions.query.LambdaQueryWrapper;
 import com.baomidou.mybatisplus.extension.service.impl.ServiceImpl;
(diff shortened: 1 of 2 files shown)


streampark-console/streampark-console-service/src/test/java/org/apache/streampark/console/core/service/SettingServiceTest.java
```diff
@@ -98,4 +98,14 @@ void checkEmailTest() {
...
     Assertions.assertEquals(result.getStatus(), 200);
   }
+
+  @Test
+  void checkDockerTest() {
+    DockerConfig dockerConfig = new DockerConfig();
+    dockerConfig.setAddress("https://hub.docker.com/v2/users/login");
+    dockerConfig.setUserName("username");
+    dockerConfig.setPassword("password");
+    dockerConfig.setNamespace("");
```
caicancai, line 108: dockerConfig.setNamespace("streampark");

caicancai (changes requested):

streampark-console/streampark-console-service/src/test/java/org/apache/streampark/console/core/service/SettingServiceTest.java
caicancai: I don't think there is any use in adding this test now, I suggest adding it together after adding the e2e test

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/SettingServiceImpl.java
```diff
@@ -145,13 +152,47 @@ public String getIngressModeDefault() {
...
+      result.setMsg("request failed, please check: " + e.getMessage());
+    }
+
     return result;
   }
 
+  private static HttpPost getHttpPost(final DockerConfig dockerConfig) {
+    HttpPost loginRequest = new HttpPost(dockerConfig.getAddress());
+    loginRequest.setHeader("Content-Type", "application/json");
+    String userLoginParam =
```
wolfboys, line 185: The username and password set in this way are invalid. We need to login for verification through the docker-java API. see this [PR](https://github.com/apache/incubator-streampark/pull/3593)

PR #3625, opened 2024-03-23
[Improve] AssertUtils add and check not null improvements

wolfboys:
[Improve] AssertUtils add and check not null improvements

Issue Number: close #xxx 

## Verifying this change

*(or)*

*(or)*


wolfboys:
cc @RocMarshal PTAL, thx

streampark-common/src/main/java/org/apache/streampark/common/util/ExceptionUtils.java
RocMarshal, line 1: Does this change seem unrelated to the current ISSUE title?

Perhaps it would be better to record it separately in an ISSUE.

streampark-common/src/main/scala/org/apache/streampark/common/util/AssertUtils.scala
```diff
@@ -0,0 +1,282 @@
...
+import org.apache.streampark.common.util.Utils.isEmpty
+
+import javax.annotation.Nullable
+
+import java.util
+
+import scala.collection.convert.ImplicitConversions._
+
+/** @since 2.2.0 */
+object AssertUtils {
```
RocMarshal, line 28: How about 
`AssertUtils` -> `Premises` ?

RocMarshal (approved):
Hi, @wolfboys  Nice pr!. Thank you for your hard work.
LGTM on the whole.

Just a few of comments. PTAL in your free time~

streampark-common/src/main/scala/org/apache/streampark/common/util/AssertUtils.scala
wolfboys: > How about `AssertUtils` -> `Premises` ?

oh, I think AssertUtils is a nice name, e.g: the name in spring is [Assert](https://github.com/spring-projects/spring-framework/blob/5.1.x/spring-core/src/main/java/org/springframework/util/Assert.java)

caicancai (approved):
LGTM

PR #3636, opened 2024-03-27
[ISSUES-3627] doSetFlinkConf and doSetVersion are changed in sequence

hackallan:

Issue Number: close #xxx 

## Verifying this change

*(or)*

*(or)*

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -21,6 +21,8 @@ on:
   push:
     branches:
       - dev
+      - release-*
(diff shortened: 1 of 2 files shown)


hackallan:
doSetFlinkConf reads the getVersionOfFirst() method without a version variable value. The reason is that the create method in the FlinkEnvServiceImpl class is caused when the doSetVersion method is called after doSetFlinkConf, so change the order

wolfboys (approved):
LGTM

PR #3615, opened 2024-03-14
[Improve] mybatis-plus updateStrategy improvements

VampireAchao:
## What changes were proposed in this pull request

Issue Number: close #3607 

## Brief change log

- [x] Updated the global `updateStrategy` setting in MyBatis-Plus.
- [x] Removed the `@TableField(updateStrategy = FieldStrategy.IGNORED)` annotations across all entity classes.
- [x] Enumerated and verified all affected update operations to ensure integrity and prevent issues.

## Verifying this change

This change is a trivial rework / code cleanup without any test coverage.

*no* 

This change is already covered by existing tests, such as *(please describe tests)*.

*no*

This change added tests and can be verified as follows:

no

## Does this pull request potentially affect one of the following parts
 - Dependencies (does it add or upgrade a dependency): yes
 - add cn.hutool:hutool-core:5.8.27

PR #3655, opened 2024-04-04
Revert [Improve] mybatis-plus updateStrategy improvements

wolfboys:
Reverts apache/incubator-streampark#3615

While we have provided some testcase, many test cases are missing for the update scenarios. After my testing and confirmation, it is found that there are many bugs impacting a large area. I suggest rolling back this Pull Request. Perhaps we should reevaluate our approach, for instance, considering replacing MyBatis with Spring Data JPA in the future? 🐛🔄🤔


caicancai (approved):
LGTM. Good job

PR #3658, opened 2024-04-07
Fixed the issue that SSO could not jump when debugging the front-end …

yangzehan:
修复sso在本地调试前后端分离时无法跳转

当前本地运行streampark时是前后端分离的形式，但sso登录按钮所携带的链接是相对路径/sso/signin,前端会访问localhost:10001/sso/signin并不会访问到后端localhost:10000/sso/signin所以会出现无法跳转到认证页面的现象。
image
我在配置文件中加入了BASE_ADDRESS变量指向后端地址与SSO_LOGIN_PATH拼接于是页面可以正确跳转到认证页面

Fixed the issue that SSO could not jump when debugging the front-end and backend separation locally

When StreamPark is run locally, it is in the form of front-end and back-end separation, but the link carried by the SSO login button is the relative path /sso/signin, and the front-end will access localhost:10001/sso/signin and will not access the back-end localhost:10000/sso/signin, so it will not be able to jump to the authentication page.

I added BASE_ADDRESS variable to the config file to point to the backend address and concatenate it with the SSO_LOGIN_PATH, so that the page can jump to the authentication page correctly

### DIFF
--- a/streampark-console/streampark-console-webapp/.env.development
+++ b/streampark-console/streampark-console-webapp/.env.development
@@ -20,6 +20,8 @@ VITE_PUBLIC_PATH=/
 # Please note that no line breaks
 VITE_PROXY=[["/basic-api","http://localhost:10000"]]
 
+VITE_BASE_ADDRESS=http://localhost:10000
+
 # Delete console
 VITE_DROP_CONSOLE=false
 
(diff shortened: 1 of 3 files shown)


caicancai:
@saLeox Can you help with the review, thank you

wangsizhu0504 (approved):

PR #3659, opened 2024-04-08
[Bug][Flink-Submit] Fixed security issues with submitting external applications to manipulate the Jvm

zhilinli123:
issues：
Many frameworks restrict the passing of external parameters, and it is possible for the user to accidentally forget to pass them in. This is an actual problem that I have encountered, and it is a dangerous operation that has caused my streampark service jvm to fail directly in the past

```
  public static void main(String[] args) throws Exception {
        if (args.length < 1) {
            logger.error("Missing configuration file!");
            System.exit(1);
        } 

    }
```
### Before repair
image

### After repair

image


wolfboys:
Thanks for your contribution, I will review it later

zhilinli123:
PTAL: @wolfboys  thanks !

wolfboys (approved):
LGTM

PR #3670, opened 2024-04-16
[Improve] Added locale picker to login page

wangsizhu0504:

image


wangsizhu0504:
@wolfboys I've added language switching to the login page, please check if it's available!

wolfboys (approved):
LGTM

PR #3676, opened 2024-04-22
[Fix][Bug] Fix docker config naming problem

caicancai:

Fix userName to username

## Verifying this change

*(or)*

*(or)*

### DIFF
--- a/streampark-console/streampark-console-webapp/src/views/setting/system/config.ts
+++ b/streampark-console/streampark-console-webapp/src/views/setting/system/config.ts
@@ -47,12 +47,12 @@ export const settingFormSchema: SettingForm = {
       required: true,
     },
     {
-      field: 'userName',
-      label: t('setting.system.docker.userName.label'),
-      helpMessage: t('setting.system.docker.userName.desc'),
+      field: 'username',
+      label: t('setting.system.docker.username.label'),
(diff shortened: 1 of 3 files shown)


streampark-console/streampark-console-webapp/src/views/setting/system/config.ts
```diff
@@ -47,12 +47,12 @@ export const settingFormSchema: SettingForm = {
       required: true,
     },
     {
-      field: 'userName',
-      label: t('setting.system.docker.userName.label'),
-      helpMessage: t('setting.system.docker.userName.desc'),
+      field: 'username',
+      label: t('setting.system.docker.username.label'),
```
wangsizhu0504, line 51: Here you need to synchronize the key field in the internationalization file, which is located at streampark-console/streampark-console-webapp/src/locales/lang/en/setting/system.ts (similar for zh_CN).

wangsizhu0504 (changes requested):

streampark-console/streampark-console-webapp/src/views/setting/system/config.ts
caicancai: done

wangsizhu0504 (commented):
LGTM!

wangsizhu0504 (approved):
LGTM！

PR #3705, opened 2024-05-11
[Feature] Optimize log printing and return in GlobalExceptionHandler

Mrart:
## What changes were proposed in this pull request

Issue Number: close #3704 
Optimize log printing and return in GlobalExceptionHandler

## Brief change log

Optimize log printing and return in GlobalExceptionHandler
## Verifying this change

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)* yes

## Does this pull request potentially affect one of the following parts
 - Dependencies (does it add or upgrade a dependency): (yes )

PR #3707, opened 2024-05-13
[Imporve] Optimize log printing in GlobalExceptionHandler

caicancai:

Issue Number: close #xxx 

## Verifying this change

*(or)*

*(or)*

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/handler/GlobalExceptionHandler.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/handler/GlobalExceptionHandler.java
@@ -59,7 +59,7 @@ public class GlobalExceptionHandler {
   @ExceptionHandler(value = HttpRequestMethodNotSupportedException.class)
   @ResponseStatus(HttpStatus.INTERNAL_SERVER_ERROR)
   public RestResponse handleException(HttpRequestMethodNotSupportedException e) {
-    log.error("not supported request method，exception：{}", e.getMessage());
+    log.error("not supported request method，exception：{}", e);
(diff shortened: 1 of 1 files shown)


streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/handler/GlobalExceptionHandler.java
```diff
@@ -113,6 +113,6 @@ public RestResponse handleConstraintViolationException(ConstraintViolationExcept
   @ExceptionHandler(value = UnauthorizedException.class)
   @ResponseStatus(HttpStatus.FORBIDDEN)
   public void handleUnauthorizedException(Exception e) {
-    log.info("Permission denied，{}", e.getMessage());
+    log.error("Permission denied，{}", e.getMessage());
```
wolfboys: If e is NullPointerException, then e.getMessage will throw nullPointExp...  You can test this:

```
  @Test
  public void testException() {
    try {
      throw new NullPointerException();
    } catch (Exception e) {
      // e.getMessage will be throw nullExp....
      System.out.printf(e.getMessage());
    }
  }
```

wolfboys:
There are some minor issues here, You can refer to this [PR](https://github.com/apache/incubator-streampark/pull/3705)

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/handler/GlobalExceptionHandler.java
caicancai: done

VampireAchao (approved):
LGTM

PR #3721, opened 2024-05-23
[Improve] Improve error log

caicancai:

Improve error log

## Verifying this change

*(or)*

*(or)*


caicancai:
I have checked all the problems with service/impl. Currently, there are only these logs that I think need to be improved.
I think there is no need to add error_log after throwing exception

VampireAchao (approved):
LGTM

PR #3759, opened 2024-06-15
[Improve] h2 dialect improvement

### LINKED ISSUE #3762: [Improve] h2 dialect improvement
### Search before asking

### Description

apache StreamPark project supports H2 as a dialect for the convenience of users to experience quickly, but doesn't persist data; the data is stored in memory, so once the platform is restarted, the data is lost. H2 itself supports data persistence, so we need to support H2 data persistence.

### Usage Scenario

_No response_

### Related issues

_No response_

### Code of Conduct

wolfboys:
[Improve] h2 dialect improvement

Issue Number: close #3762

## Verifying this change

*(or)*

*(or)*

### DIFF
--- a/streampark-console/streampark-console-service/src/main/assembly/conf/config.yaml
+++ b/streampark-console/streampark-console-service/src/main/assembly/conf/config.yaml
@@ -35,7 +35,7 @@ server:
 # system database, default h2, mysql|pgsql|h2
 datasource:
   dialect: h2  #h2, mysql, pgsql
-  h2-data-dir:  # if datasource.dialect is h2 ,you can configure the data dir
+  h2-data-dir: ~/streampark/h2-data # if datasource.dialect is h2 ,you can configure the data dir
   # if datasource.dialect is mysql or pgsql, you need to configure the following connection information
   # mysql/postgresql/h2 connect user
   username:
--- a/streampark-console/streampark-console-service/src/main/assembly/conf/config.yaml
+++ b/streampark-console/streampark-console-service/src/main/assembly/conf/config.yaml
@@ -35,10 +35,11 @@ server:
 # system database, default h2, mysql|pgsql|h2
 datasource:
   dialect: h2  #h2, mysql, pgsql
+  h2-data-dir:  # if datasource.dialect is h2 ,you can configure the data dir
   # if datasource.dialect is mysql or pgsql, you need to configure the following connection information
-  # mysql/postgresql connect user
+  # mysql/postgresql/h2 connect user
   username:
-  # mysql/postgresql connect password
+  # mysql/postgresql/h2 connect password
   password:
   # mysql/postgresql connect jdbcURL
   # mysql example: datasource.url: jdbc:mysql://localhost:3306/streampark?useUnicode=true&characterEncoding=UTF-8&useJDBCCompliantTimezoneShift=true&useLegacyDatetimeCode=false&serverTimezone=GMT%2B8
--- a/streampark-console/streampark-console-service/src/main/assembly/conf/config.yaml
+++ b/streampark-console/streampark-console-service/src/main/assembly/conf/config.yaml
@@ -35,7 +35,7 @@ server:
 # system database, default h2, mysql|pgsql|h2
 datasource:
   dialect: h2  #h2, mysql, pgsql
-  h2-data-dir: ~/streampark/h2-data # if datasource.dialect is h2 ,you can configure the data dir
+  h2-data-dir: ~/streampark/h2-data # if datasource.dialect is h2, you can configure the data dir
   # if datasource.dialect is mysql or pgsql, you need to configure the following connection information
   # mysql/postgresql/h2 connect user
   username:
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/config/SpringProperties.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/config/SpringProperties.java
@@ -87,17 +87,38 @@ public class SpringProperties {
         springConfig.put("spring.datasource.driver-class-name", "org.postgresql.Driver");
         break;
       case "h2":
-        springConfig.put("spring.datasource.driver-class-name", "org.h2.Driver");
-        springConfig.put("spring.datasource.username", "sa");
-        springConfig.put("spring.datasource.password", "sa");
+        String h2Dir = userConfig.getProperty("datasource.h2-data-dir", null);
+
+        if (StringUtils.isBlank(h2Dir)) {
+          h2Dir = "~/streampark/metadata";
+        } else {
+          h2Dir += h2Dir.endsWith("/") ? "metadata" : "/metadata";
+        }
+
         springConfig.put(
             "spring.datasource.url",
-            "jdbc:h2:mem:streampark;MODE=MySQL;DB_CLOSE_DELAY=-1;DATABASE_TO_LOWER=true;INIT=runscript from 'classpath:db/schema-h2.sql'");
+            String.format(
+                "jdbc:h2:file:%s;MODE=MySQL;DB_CLOSE_DELAY=-1;DATABASE_TO_LOWER=true;INIT=runscript from 'classpath:db/schema-h2.sql'",
+                h2Dir));
+
+        String userName = userConfig.getProperty("spring.datasource.username", "admin");
+        String password = userConfig.getProperty("spring.datasource.password", "streampark");
+
+        springConfig.put("spring.jpa.database-platform", "org.hibernate.dialect.H2Dialect");
+        springConfig.put("spring.datasource.driver-class-name", "org.h2.Driver");
+        springConfig.put("spring.datasource.username", userName);
+        springConfig.put("spring.datasource.password", password);
         springConfig.put("spring.sql.init.data-locations", "classpath:db/data-h2.sql");
         springConfig.put("spring.sql.init.continue-on-error", "true");
-        springConfig.put("spring.sql.init.username", "sa");
-        springConfig.put("spring.sql.init.password", "sa");
+        springConfig.put("spring.sql.init.username", userName);
+        springConfig.put("spring.sql.init.password", password);
         springConfig.put("spring.sql.init.mode", "always");
+
+        // h2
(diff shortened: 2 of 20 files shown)


wolfboys:
cc @GOODBOY008 @SbloodyS PTAL, thx, I have tested it and it passed the test, I will update the doc when version 2.1.5 is released.

streampark-console/streampark-console-service/src/main/assembly/conf/config.yaml
```diff
@@ -35,10 +35,11 @@ server:
 # system database, default h2, mysql|pgsql|h2
 datasource:
   dialect: h2  #h2, mysql, pgsql
+  h2-data-dir:  # if datasource.dialect is h2 ,you can configure the data dir
```
SbloodyS:
```suggestion
  h2-data-dir:  /tmp/streampark/h2-data # if datasource.dialect is h2 ,you can configure the data dir
```

I think it's better to set a default value for it. And it should be the same as default value.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/config/SpringProperties.java
```diff
@@ -87,17 +87,37 @@ private static void dataSourceConfig(Properties userConfig, Properties springCon
         springConfig.put("spring.datasource.driver-class-name", "org.postgresql.Driver");
         break;
       case "h2":
-        springConfig.put("spring.datasource.driver-class-name", "org.h2.Driver");
-        springConfig.put("spring.datasource.username", "sa");
-        springConfig.put("spring.datasource.password", "sa");
+        String h2Dir = userConfig.getProperty("datasource.h2-data-dir", null);
+        if (StringUtils.isBlank(h2Dir)) {
+          h2Dir = System.getProperty("user.home", "~") + "/streampark/metadata";
```
SbloodyS: I think we should directly using `h2Dir`  value so that can reduce the user's cost of use.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/config/SpringProperties.java
```diff
@@ -87,17 +87,37 @@ private static void dataSourceConfig(Properties userConfig, Properties springCon
...
+        } else {
+          h2Dir += h2Dir.endsWith("/") ? "metadata" : "/metadata";
```
SbloodyS: the same as `/tmp/streampark/h2-data`

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/CheckpointProcessor.java
```diff
@@ -131,6 +131,18 @@ private void process(Application application, @Nonnull CheckPoints.CheckPoint ch
     }
   }
 
+  // issue: https://github.com/apache/incubator-streampark/issues/3749
```
SbloodyS: We should avoid this meaningless comment since we should do that in commit message and pr.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/config/SpringProperties.java
```diff
@@ -87,17 +87,37 @@ private static void dataSourceConfig(Properties userConfig, Properties springCon
         springConfig.put("spring.datasource.driver-class-name", "org.postgresql.Driver");
         break;
       case "h2":
-        springConfig.put("spring.datasource.driver-class-name", "org.h2.Driver");
-        springConfig.put("spring.datasource.username", "sa");
-        springConfig.put("spring.datasource.password", "sa");
+        String h2Dir = userConfig.getProperty("datasource.h2-data-dir", null);
```
SbloodyS: Is it better to use `h2DataDir` instead of `h2Dir` which is more intuitive.

streampark-console/streampark-console-service/src/main/assembly/conf/config.yaml
wolfboys: Users may not necessarily have permission for this directory. So I think it is not suitable to provide a default value.

SbloodyS: On Linux systems, all users have permissions on the `/tmp` directory by default since it's default permissions is `777`.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/config/SpringProperties.java
wolfboys: This parameter is read from config.yaml. h2DataDir naming style has a minor issue.

streampark-console/streampark-console-service/src/main/assembly/conf/config.yaml
wolfboys: > On Linux systems, all users have permissions on the `/tmp` directory by default since it's default permissions is `777`.

"/tmp" is a system temporary directory. I am concerned that it might get cleaned, as some users have scripts that regularly clean up directories and files in "/tmp." Therefore, it is not recommended to place it in "/tmp." It is better for users to manually specify the path using "h2dir" or have the program directly locate the directory under the user's home directory. What do you think？

SbloodyS: I don't think leaving the default value blank is a good option since we using `h2` as default database. Providing the default value allows users to use it directly without the need to configure it. If users use it in a production environment, we can suggest that users change to other values in the documentation. WDYT?

wolfboys: > I don't think leaving the default value blank is a good option since we using `h2` as default database. Providing the default value allows users to use it directly without the need to configure it. If users use it in a production environment, we can suggest that users change to other values in the documentation. WDYT?

We agree with your opinion that default values should be set, but "/tmp/xxx" is not a good solution. Perhaps `~/streampark/metadata` is a good idea, using `~` will automatically locate the user directory.

VampireAchao: If it is for demonstration purposes, I prefer the /tmp directory. However, we need to consider users who might use it to run lightweight jobs and need persistence. In this case, the /tmp directory might not be suitable. I prefer the /data/streampark/ directory. This not only avoids the habitual cleaning of the /tmp directory that might lead to data loss but also aligns with the conventions of some projects, such as:

Docker: By default, Docker containers use /data for storing persistent data.
Elasticsearch: The default data directory for Elasticsearch is /var/lib/elasticsearch which often gets customized to /data/elasticsearch.
MongoDB: Uses /data/db as the default directory for storing data.
Prometheus: The default data directory is /var/lib/prometheus, and many setups prefer using /data/prometheus.
These examples demonstrate a common practice of using the /data directory for persistent storage, ensuring stability and avoiding data loss due to temporary directory cleaning.

What do you think?

VampireAchao: > > I don't think leaving the default value blank is a good option since we using `h2` as default database. Providing the default value allows users to use it directly without the need to configure it. If users use it in a production environment, we can suggest that users change to other values in the documentation. WDYT?
> 
> We agree with your opinion that default values should be set, but "/tmp/xxx" is not a good solution. Perhaps `~/streampark/metadata` is a good idea, using `~` will automatically locate the user directory.

Using the ~ directory might lead to data loss when users change or leave and their accounts get deleted. Of course, we can also inform users about this potential issue in the documentation.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/config/SpringProperties.java
wolfboys: > Is it better to use `h2DataDir` instead of `h2Dir` which is more intuitive.

done

streampark-console/streampark-console-service/src/main/assembly/conf/config.yaml
```diff
@@ -35,10 +35,11 @@ server:
 # system database, default h2, mysql|pgsql|h2
 datasource:
   dialect: h2  #h2, mysql, pgsql
+  h2-data-dir: ~/streampark/h2-data # if datasource.dialect is h2 ,you can configure the data dir
```
SbloodyS:
```suggestion
  h2-data-dir: ~/streampark/h2-data # if datasource.dialect is h2, you can configure the data dir
```

streampark-console/streampark-console-service/src/main/assembly/conf/config.yaml
SbloodyS: > We agree with your opinion that default values should be set, but "/tmp/xxx" is not a good solution. Perhaps `~/streampark/metadata` is a good idea, using `~` will automatically locate the user directory.

Using `~/streampark/metadata` is good to me.

wolfboys: > > We agree with your opinion that default values should be set, but "/tmp/xxx" is not a good solution. Perhaps `~/streampark/metadata` is a good idea, using `~` will automatically locate the user directory.
> 
> Using `~/streampark/metadata` is good to me.

Thanks, updated~

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/config/SpringProperties.java
wolfboys: > I think we should directly using `h2Dir` value so that can reduce the user's cost of use.

updated.

streampark-console/streampark-console-service/src/main/assembly/conf/config.yaml
wolfboys: updated

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/CheckpointProcessor.java
wolfboys: > We should avoid this meaningless comment since we should do that in commit message and pr.

Thanks, updated

SbloodyS (approved):
LGTM.

PR #3764, opened 2024-06-16
[Imporve] Set required_approving_review_count to 2

caicancai:

Issue Number: close #xxx 

## Verifying this change

*(or)*

*(or)*


wolfboys:
hi caicancai:

Perhaps other projects already have such rules, but each project has its way of collaboration. Introducing rules is to solve problems and bring convenience, not to create obstacles. Currently, there are relatively few committers and PPMC members for Project StreamPark; these committers and PPMC members are not very active either. So I think it's not applicable now.

PR #3806, opened 2024-06-25
[improve] code polish : log, logic, exception etc

GOODBOY008:
[improve] code polish : log, logic, exception etc

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/domain/RestResponse.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/domain/RestResponse.java
@@ -35,9 +35,7 @@ public class RestResponse extends HashMap<String, Object> {
   private static final long serialVersionUID = 1L;
 
   public static RestResponse success(Object data) {
-    RestResponse resp = new RestResponse();
-    resp.put(STATUS_KEY, STATUS_SUCCESS);
-    resp.put(CODE_KEY, ResponseCode.CODE_SUCCESS);
+    RestResponse resp = success();
     resp.put(DATA_KEY, data);
     return resp;
   }
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/controller/ApplicationBuildPipelineController.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/controller/ApplicationBuildPipelineController.java
@@ -49,13 +49,9 @@ public class ApplicationBuildPipelineController {
   @PermissionScope(app = "#appId")
   @PostMapping(value = "build")
   @RequiresPermissions("app:create")
-  public RestResponse buildApplication(Long appId, boolean forceBuild) {
-    try {
-      boolean actionResult = appBuildPipeService.buildApplication(appId, forceBuild);
-      return RestResponse.success(actionResult);
-    } catch (Exception e) {
-      return RestResponse.success(false).message(e.getMessage());
-    }
+  public RestResponse buildApplication(Long appId, boolean forceBuild) throws Exception {
+    boolean actionResult = appBuildPipeService.buildApplication(appId, forceBuild);
+    return RestResponse.success(actionResult);
   }
 
   /**
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/controller/ApplicationController.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/controller/ApplicationController.java
@@ -197,13 +197,9 @@ public class ApplicationController {
   @PermissionScope(app = "#app.id", team = "#app.teamId")
   @PostMapping(value = "start")
   @RequiresPermissions("app:start")
-  public RestResponse start(@Parameter(hidden = true) Application app) {
-    try {
-      applicationActionService.start(app, false);
-      return RestResponse.success(true);
-    } catch (Exception e) {
-      return RestResponse.success(false).message(e.getMessage());
-    }
+  public RestResponse start(@Parameter(hidden = true) Application app) throws Exception {
+    applicationActionService.start(app, false);
+    return RestResponse.success(true);
   }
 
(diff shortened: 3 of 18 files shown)


streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/controller/ApplicationBuildPipelineController.java
```diff
@@ -49,13 +49,9 @@ public class ApplicationBuildPipelineController {
   @PermissionScope(app = "#appId")
   @PostMapping(value = "build")
   @RequiresPermissions("app:create")
-  public RestResponse buildApplication(Long appId, boolean forceBuild) {
-    try {
-      boolean actionResult = appBuildPipeService.buildApplication(appId, forceBuild);
-      return RestResponse.success(actionResult);
-    } catch (Exception e) {
-      return RestResponse.success(false).message(e.getMessage());
```
wolfboys, line 57: The FE will logical processing based on the controller response. If an exception is thrown here, it will cause a bug in the front end.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/controller/ApplicationController.java
```diff
@@ -197,13 +197,9 @@ public RestResponse checkStart(Application app) {
   @PermissionScope(app = "#app.id", team = "#app.teamId")
   @PostMapping(value = "start")
   @RequiresPermissions("app:start")
-  public RestResponse start(@Parameter(hidden = true) Application app) {
-    try {
-      applicationActionService.start(app, false);
-      return RestResponse.success(true);
-    } catch (Exception e) {
-      return RestResponse.success(false).message(e.getMessage());
```
wolfboys, line 205: I‘m worried that this will cause bugs on the front end.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/controller/FlinkEnvController.java
```diff
@@ -53,12 +52,8 @@ public RestResponse check(FlinkEnv version) {
   }
 
   @PostMapping("create")
-  public RestResponse create(FlinkEnv version) {
-    try {
-      flinkEnvService.create(version);
-    } catch (Exception e) {
-      throw new ApiDetailException(e);
```
wolfboys, line 60: ApiDetailException is a special exception. The front end will display the exception through an alert pop-up window. If it is removed, the exception message will not be popped up to notify the user.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/controller/SettingController.java
```diff
@@ -110,12 +111,8 @@ public RestResponse updateEmail(SenderEmail senderEmail) {
   }
 
   @PostMapping("check/hadoop")
-  public RestResponse checkHadoop() {
-    try {
-      HadoopUtils.hdfs().getStatus();
-      return RestResponse.success(true);
-    } catch (Exception e) {
-      return RestResponse.success(false).message(e.getMessage());
```
wolfboys, line 118: This may cause bug in the FE

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/domain/RestResponse.java
```diff
@@ -35,9 +35,7 @@ public class RestResponse extends HashMap<String, Object> {
   private static final long serialVersionUID = 1L;
 
   public static RestResponse success(Object data) {
-    RestResponse resp = new RestResponse();
-    resp.put(STATUS_KEY, STATUS_SUCCESS);
-    resp.put(CODE_KEY, ResponseCode.CODE_SUCCESS);
+    RestResponse resp = success();
```
wolfboys, line 38: good job

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/application/impl/ApplicationActionServiceImpl.java
```diff
@@ -796,7 +791,7 @@ private void doAbort(Long id) {
           YarnClient yarnClient = HadoopUtils.yarnClient();
           yarnClient.killApplication(applications.get(0).getApplicationId());
         }
-      } catch (Exception exception) {
+      } catch (Exception ignored) {
```
wolfboys: The exception should be recorded and not ignored

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/controller/ApplicationBuildPipelineController.java
GOODBOY008: @wolfboys `org.apache.streampark.console.base.handler.GlobalExceptionHandler` We have global exception handler,so there is no need to surround with try-catch.

wolfboys: Get it, Thanks for your clarification

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/controller/FlinkEnvController.java
wolfboys: How to show this exception info to the [front end](https://github.com/apache/incubator-streampark/blob/44f67364979d2675fd3a2c0205ebe4223d24f672/streampark-console/streampark-console-webapp/src/views/flink/cluster/Add.vue#L55) needs further improvement

image

GOODBOY008: @wolfboys Updated.

wolfboys (approved):
LGTM

PR #3877, opened 2024-07-17
[Hotfix] The MySQL Connector/J is incompatible with the Apache 2.0 license.

zhuangchong:

The binary package lib directory contains `mysql-connector-java-8.0.27.jar`

Remove this jar because MySQL Connector/J is not compatible with Apache 2.0 licenses.

image

## Verifying this change

*(or)*

*(or)*


pom.xml
```diff
@@ -240,6 +240,7 @@
                 <groupId>mysql</groupId>
                 <artifactId>mysql-connector-java</artifactId>
                 <version>${mysql.version}</version>
+                <scope>test</scope>
```
SbloodyS, line 243: We should add this to the docs to tell users downloading mysql jar locally.

zhuangchong: There are instructions in the deployment document (https://streampark.apache.org/docs/platform/platform-build#introduce-mysql-dependency-package). At the same time, the instructions will be checked when starting the project (https://github.com/apache/incubator-streampark/blob/f9421f6bbc162200cd9accdc25ac2bd43e78e889/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/config/SpringProperties.java#L56-L69)

SbloodyS: I see. Thanks for clarification.

wolfboys (approved):
LGTM

PR #3901, opened 2024-07-20
[Improve] spark-submit some improvements

wolfboys:

Issue Number: close #xxx 

## Verifying this change

*(or)*

*(or)*

### DIFF
--- a/streampark-common/src/main/java/org/apache/streampark/common/enums/SparkDevelopmentMode.java
+++ b/streampark-common/src/main/java/org/apache/streampark/common/enums/SparkDevelopmentMode.java
@@ -48,7 +48,7 @@ public enum SparkDevelopmentMode {
      * @return The parsed flink deployment mode.
      */
     @Nonnull
-    public static SparkDevelopmentMode of(@Nullable Integer value) {
+    public static SparkDevelopmentMode valueOf(@Nullable Integer value) {
         for (SparkDevelopmentMode flinkDevelopmentMode : values()) {
             if (flinkDevelopmentMode.mode.equals(value)) {
                 return flinkDevelopmentMode;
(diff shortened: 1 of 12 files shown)


streampark-common/src/main/java/org/apache/streampark/common/enums/SparkDevelopmentMode.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    /**
+     * Try to resolve the mode value into {@link SparkDevelopmentMode}.
+     *
+     * @param value The mode value of potential flink deployment mode.
+     * @return The parsed flink deployment mode.
+     */
+    @Nonnull
+    public static SparkDevelopmentMode of(@Nullable Integer value) {
```
SbloodyS:
```suggestion
    public static SparkDevelopmentMode valusOf(@Nullable Integer value) {
```

It's better to use an explicit naming method.

streampark-common/src/main/java/org/apache/streampark/common/enums/SparkDevelopmentMode.java
```diff
@@ -0,0 +1,65 @@
...
+            }
+        }
+        return SparkDevelopmentMode.UNKNOWN;
+    }
+
+    /** Get the mode value of the current {@link SparkDevelopmentMode} enum. */
+    @Nonnull
+    public Integer getMode() {
+        return mode;
+    }
```
SbloodyS: Using `@Getter` is better instead of this.

streampark-common/src/main/java/org/apache/streampark/common/enums/SparkDevelopmentMode.java
wolfboys: done

streampark-common/src/main/java/org/apache/streampark/common/enums/SparkDevelopmentMode.java
wolfboys: unused method, removed

caicancai (approved):

PR #3908, opened 2024-07-22
[Bug] Correct spark app submitting process

lenoxzhao:

Issue Number: close #xxx 

## Brief change log

1. For `java.util.regex.Pattern`, `matcher.group()` should be used after matching method like `matches()` and `find()`, or an exception will be thrown.
2. `spark-submit --version` prints spark version and scala version, so `StringUtils.isBlank(sparkVersion)` is need to ensure `version`  be assigned the correct value.image
3. `SparkHandle.State` represents the state of job rather than the submission. So if we want to return the result after the submission finished, we should not use `handle.getState.isFinal`.

## Verifying this change

*(or)*

*(or)*

### DIFF
--- a/streampark-common/src/main/scala/org/apache/streampark/common/conf/SparkVersion.scala
+++ b/streampark-common/src/main/scala/org/apache/streampark/common/conf/SparkVersion.scala
@@ -43,7 +43,7 @@ class SparkVersion(val sparkHome: String) extends Serializable with Logger {
     if (matcher.matches()) {
       matcher.group(1)
     } else {
-      "2.12"
+      throw new IllegalArgumentException(s"[StreamPark] can not found scala version in $sparkCoreJar")
     }
   }
 
--- a/streampark-common/src/main/scala/org/apache/streampark/common/conf/SparkVersion.scala
+++ b/streampark-common/src/main/scala/org/apache/streampark/common/conf/SparkVersion.scala
@@ -20,6 +20,8 @@ package org.apache.streampark.common.conf
 import org.apache.streampark.common.util.{CommandUtils, Logger}
 import org.apache.streampark.common.util.Implicits._
 
+import org.apache.commons.lang3.StringUtils
+
 import java.io.File
 import java.net.URL
 import java.util.function.Consumer
@@ -36,7 +38,14 @@ class SparkVersion(val sparkHome: String) extends Serializable with Logger {
 
   private[this] lazy val SPARK_SCALA_VERSION_PATTERN = Pattern.compile("^spark-core_(.*)-[0-9].*.jar$")
 
-  lazy val scalaVersion: String = SPARK_SCALA_VERSION_PATTERN.matcher(sparkCoreJar.getName).group(1)
+  lazy val scalaVersion: String = {
+    val matcher = SPARK_SCALA_VERSION_PATTERN.matcher(sparkCoreJar.getName)
+    if (matcher.matches()) {
+      matcher.group(1)
+    } else {
+      "2.12"
+    }
+  }
 
   lazy val sparkCoreJar: File = {
     val distJar = sparkLib.listFiles().filter(_.getName.matches("spark-core.*\\.jar"))
@@ -98,7 +107,7 @@ class SparkVersion(val sparkHome: String) extends Serializable with Logger {
         override def accept(out: String): Unit = {
           buffer.append(out).append("\n")
           val matcher = SPARK_VERSION_PATTERN.matcher(out)
-          if (matcher.find) {
+          if (matcher.find && StringUtils.isBlank(sparkVersion)) {
             sparkVersion = matcher.group(2)
           }
         }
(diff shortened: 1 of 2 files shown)


streampark-common/src/main/scala/org/apache/streampark/common/conf/SparkVersion.scala
```diff
@@ -36,7 +38,14 @@ class SparkVersion(val sparkHome: String) extends Serializable with Logger {
 
   private[this] lazy val SPARK_SCALA_VERSION_PATTERN = Pattern.compile("^spark-core_(.*)-[0-9].*.jar$")
 
-  lazy val scalaVersion: String = SPARK_SCALA_VERSION_PATTERN.matcher(sparkCoreJar.getName).group(1)
+  lazy val scalaVersion: String = {
```
wolfboys: It would be better to throw an exception if the Scala version cannot be extracted.
```
  lazy val scalaVersion: String = {
    val matcher = SPARK_SCALA_VERSION_PATTERN.matcher(sparkCoreJar.getName)
    if (matcher.find()) {
      matcher.group(1)
    } else {
      throw new IllegalArgumentException(s"[StreamPark] can not found scala version in $sparkCoreJar")
    }
  }
```

wolfboys, line 41: If the Scala version number cannot be obtained, an exception should be thrown.
```
 lazy val scalaVersion: String = {
    val matcher = SPARK_SCALA_VERSION_PATTERN.matcher(sparkCoreJar.getName)
    if (matcher.find()) {
      matcher.group(1)
    } else {
      throw new IllegalArgumentException(s"[StreamPark] can not found scala version in $sparkCoreJar")
    }
  }

```

streampark-common/src/main/scala/org/apache/streampark/common/conf/SparkVersion.scala
```diff
@@ -98,7 +107,7 @@ class SparkVersion(val sparkHome: String) extends Serializable with Logger {
         override def accept(out: String): Unit = {
           buffer.append(out).append("\n")
           val matcher = SPARK_VERSION_PATTERN.matcher(out)
-          if (matcher.find) {
+          if (matcher.find && StringUtils.isBlank(sparkVersion)) {
```
wolfboys: Why is it necessary to check if sparkVersion is blank?

lenoxzhao: Because the pattern is `Pattern.compile("(version) (\\d+\\.\\d+\\.\\d+)")` so the output of `spark-submit --version` will be matched twice, first time to `version 3.1.2` which corresponds to the Spark's  version, and the second time to `version 2.12.10` which corresponds to the Scala's version. So if sparkVersion is not blank, it means that it has already been assigned with Spark's version value. If we go into the if statement again, sparkVersion will be overwritten by Scala's version value.
1211721648224_ pic

streampark-common/src/main/scala/org/apache/streampark/common/conf/SparkVersion.scala
lenoxzhao: done

streampark-spark/streampark-spark-client/streampark-spark-client-core/src/main/scala/org/apache/streampark/spark/client/impl/YarnClient.scala
```diff
@@ -26,15 +26,26 @@ import org.apache.streampark.spark.client.bean._
...
+import java.util.concurrent.{ConcurrentHashMap, CountDownLatch}
 
 import scala.util.{Failure, Success, Try}
 
 /** yarn application mode submit */
 object YarnClient extends SparkClientTrait {
 
+  private lazy val sparkHandles = new ConcurrentHashMap[String, SparkAppHandle]()
+
   override def doStop(stopRequest: StopRequest): StopResponse = {
```
wolfboys, line 38: You need to consider what to do if an exception occurs. What happens if the SparkAppHandle kill fails?
```
override def doStop(stopRequest: StopRequest): StopResponse = {
    val sparkAppHandle = sparkHandles.remove(stopRequest.jobId)
    if (sparkAppHandle != null) {
      Try(sparkAppHandle.kill()) match {
        case Success(_) =>
          logger.warn(s"[StreamPark][Spark][YarnClient] spark job: ${stopRequest.jobId} is stopped successfully.")
          StopResponse(null)
        case Failure(e) =>
          logger.error("sparkAppHandle kill failed. Try kill by yarn", e)
          yarnKill(stopRequest.jobId)
          StopResponse(null)
      }
    } else {
      yarnKill(stopRequest.jobId)
      StopResponse(null)
    }
  }

  private def yarnKill(appId: String): Unit = {
    Try(HadoopUtils.yarnClient.killApplication(ApplicationId.fromString(appId))) match {
      case Failure(e) => throw e
      case _ =>
    }
  }
```

wolfboys (approved):
LGTM

PR #3922, opened 2024-07-27
[Improve] default team improvement

wolfboys:
[Improve] default team improvement

## Change log

When a user logs in to the StreamPark platform for the first time, a pop-up window asks the user to select a team. However, a better approach would be for the system to set the default team as the user's default team if their teamId has not been set. This would enhance the user experience.

## Verifying this change

*(or)*

*(or)*


wangsizhu0504 (approved):
LGTM

PR #3923, opened 2024-07-28
[Test-CI] Refactor unify rest response.

wolfboys:

Issue Number: close #xxx 

## Verifying this change

*(or)*

*(or)*


streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/controller/ApplicationController.java
```diff
@@ -79,190 +78,187 @@ public class ApplicationController {
...
-        return RestResponse.success(true);
+        return Response.success();
     }
 
     @PostMapping("dashboard")
     @Permission(team = "#teamId")
-    public RestResponse dashboard(Long teamId) {
-        Map<String, Serializable> dashboardMap = applicationInfoService.getDashboardDataMap(teamId);
-        return RestResponse.success(dashboardMap);
+    public Response<?> dashboard(Long teamId) {
```
SbloodyS, line 112: Using generics in controller is not a good practice. A good crontroller should returns a specific bean object.

PR #3948, opened 2024-08-08
[Feature-3934] Add registy module for streampark

### LINKED ISSUE #3934: [Feature] Add registy module for streampark
### Search before asking

### Description

todo list:

## Function

## How to use

### streampark-registry-all

### streampark-registry-api

### Usage Scenario

_No response_

### Related issues

_No response_

### Code of Conduct

SbloodyS:

Issue Number: close #3934

## Verifying this change

*(or)*

*(or)*

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,6 +67,7 @@
         <module>streampark-flink</module>
         <module>streampark-spark</module>
         <module>streampark-console</module>
+        <module>streampark-registry</module>
     </modules>
 
     <scm>
@@ -118,7 +119,7 @@
         <json4s-jackson.version>4.0.6</json4s-jackson.version>
         <commons-cli.version>1.5.0</commons-cli.version>
         <commons-net.version>3.9.0</commons-net.version>
-        <commons-lang3.version>3.8.1</commons-lang3.version>
+        <commons-lang3.version>3.12.0</commons-lang3.version>
         <enumeratum.version>1.6.1</enumeratum.version>
         <assertj.version>3.23.1</assertj.version>
         <pprint.version>0.8.1</pprint.version>
@@ -146,6 +147,14 @@
         <mockito.version>3.4.6</mockito.version>
         <mockito-scala.version>1.17.14</mockito-scala.version>
         <assertj.version>3.23.1</assertj.version>
+        <testcontainer.version>1.19.3</testcontainer.version>
+        <zookeeper.version>3.8.0</zookeeper.version>
+        <curator.version>5.3.0</curator.version>
+        <snappy.version>1.1.10.1</snappy.version>
+        <mybatis-plus.version>3.5.3.1</mybatis-plus.version>
+        <oshi-core.version>6.1.1</oshi-core.version>
+        <commons-collections4.version>4.3</commons-collections4.version>
+        <hibernate-validator.version>6.2.2.Final</hibernate-validator.version>
 
         <maven.scaladoc.skip>false</maven.scaladoc.skip>
         <maven.scalastyle.skip>false</maven.scalastyle.skip>
@@ -158,6 +167,20 @@
     <dependencyManagement>
 
         <dependencies>
+            <!-- spring -->
+            <dependency>
+                <groupId>org.springframework.boot</groupId>
+                <artifactId>spring-boot-dependencies</artifactId>
+                <version>2.7.11</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+
+            <dependency>
+                <groupId>org.hibernate.validator</groupId>
+                <artifactId>hibernate-validator</artifactId>
+                <version>${hibernate-validator.version}</version>
+            </dependency>
 
             <dependency>
                 <groupId>org.apache.flink</groupId>
@@ -279,6 +302,46 @@
                 </exclusions>
             </dependency>
 
+            <!-- mybatis-plus -->
+            <dependency>
+                <groupId>com.baomidou</groupId>
+                <artifactId>mybatis-plus-boot-starter</artifactId>
+                <version>${mybatis-plus.version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>com.baomidou</groupId>
+                <artifactId>mybatis-plus</artifactId>
+                <version>${mybatis-plus.version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>com.baomidou</groupId>
+                <artifactId>mybatis-plus-annotation</artifactId>
+                <version>${mybatis-plus.version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>com.github.oshi</groupId>
+                <artifactId>oshi-core</artifactId>
+                <version>${oshi-core.version}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>org.slf4j</groupId>
+                        <artifactId>slf4j-simple</artifactId>
+                    </exclusion>
+                    <exclusion>
+                        <groupId>org.junit.jupiter</groupId>
+                        <artifactId>junit-jupiter-api</artifactId>
+                    </exclusion>
+                    <exclusion>
+                        <groupId>org.hamcrest</groupId>
+                        <artifactId>hamcrest</artifactId>
+                    </exclusion>
+                </exclusions>
+            </dependency>
+
+            <!--jackson-->
             <dependency>
                 <groupId>com.fasterxml.jackson.core</groupId>
                 <artifactId>jackson-databind</artifactId>
@@ -292,6 +355,12 @@
             </dependency>
 
             <dependency>
+                <groupId>com.fasterxml.jackson.datatype</groupId>
+                <artifactId>jackson-datatype-jsr310</artifactId>
+                <version>${jackson.version}</version>
+            </dependency>
+
+            <dependency>
                 <groupId>com.fasterxml.jackson.module</groupId>
                 <artifactId>jackson-module-scala_${scala.binary.version}</artifactId>
                 <version>${jackson.version}</version>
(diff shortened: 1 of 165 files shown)


streampark-common/src/main/java/org/apache/streampark/common/utils/UUIDUtils.java
```diff
@@ -0,0 +1,32 @@
...
+import lombok.experimental.UtilityClass;
+
+import java.util.Arrays;
+import java.util.UUID;
+
+@UtilityClass
+public class UUIDUtils {
+
+    public static long generateUUID() {
+        String uuidString = UUID.randomUUID().toString().replaceAll("[^0-9]]", "");
```
HxpSerein: According to [stackoverflow](https://stackoverflow.com/questions/15184820/how-to-generate-unique-positive-long-using-uuid), I perfer  `UUID.randomUUID().getMostSignificantBits() & Long.MAX_VALUE`.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/registry/ConsoleRegistryClient.java
```diff
@@ -0,0 +1,161 @@
...
+        while (ServerStatus.BUSY.equals(heartBeat.getServerStatus())) {
+            log.warn("Console node is BUSY: {}", heartBeat);
+            heartBeat = consoleHeartBeatTask.getHeartBeat();
+            ThreadUtils.sleep(SLEEP_TIME_MILLIS);
+        }
+
+        // remove before persist
+        registryClient.remove(consoleRegistryPath);
+        registryClient.persistEphemeral(consoleRegistryPath,
+            JSONUtils.toJsonString(consoleHeartBeatTask.getHeartBeat()));
```
HxpSerein, line 130: why not just use `heartBeat` instead of creating a new one?

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/registry/ConsoleRegistryClient.java
```diff
@@ -0,0 +1,161 @@
...
+     */
+    void registry() {
+        log.info("Console node : {} registering to registry center", consoleConfig.getConsoleAddress());
+        String consoleRegistryPath = consoleConfig.getConsoleRegistryPath();
+
+        ConsoleHeartBeat heartBeat = consoleHeartBeatTask.getHeartBeat();
+        while (ServerStatus.BUSY.equals(heartBeat.getServerStatus())) {
+            log.warn("Console node is BUSY: {}", heartBeat);
+            heartBeat = consoleHeartBeatTask.getHeartBeat();
+            ThreadUtils.sleep(SLEEP_TIME_MILLIS);
```
HxpSerein, line 123: move sleep before `getHeartBeat()`?

HxpSerein (commented):
Thank you for your efforts and hard work on this PR! I have reviewed the submitted code and found some issues. PTAL~

streampark-common/src/main/java/org/apache/streampark/common/utils/UUIDUtils.java
SbloodyS: Good point. Will fix it.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/registry/ConsoleRegistryClient.java
SbloodyS: This is for the future if new node types are to be added. eg: `alertServerHeartBeat`

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/registry/ConsoleRegistryClient.java
SbloodyS: This code is not currently valid. But it is a good optimization point. Will fix it.

streampark-common/src/main/java/org/apache/streampark/common/utils/UUIDUtils.java
SbloodyS: Since the maximum length is too long for database column, I reduced the length to 14 bits. I also added multithreaded unit tests to ensure there are no conflicts. PTAL @HxpSerein

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/registry/ConsoleRegistryClient.java
SbloodyS: Fixed.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/registry/ConsoleRegistryClient.java
HxpSerein: ok

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/registry/ConsoleRegistryClient.java
HxpSerein: ok

streampark-common/src/main/java/org/apache/streampark/common/utils/UUIDUtils.java
HxpSerein: ok

pom.xml
```diff
@@ -146,6 +147,14 @@
         <mockito.version>3.4.6</mockito.version>
         <mockito-scala.version>1.17.14</mockito-scala.version>
         <assertj.version>3.23.1</assertj.version>
+        <testcontainer.version>1.19.3</testcontainer.version>
+        <zookeeper.version>3.8.0</zookeeper.version>
```
wolfboys, line 151: Please make sure to declare these dependencies in the [LICENSE](https://github.com/apache/incubator-streampark/blob/dev/dist-material/release-docs/LICENSE)

wolfboys, line 151: Import ZooKeeper may result in additional dependencies, potentially causing dependency conflicts. Can you confirm this?

SbloodyS: These two dependency only using in test scope. So there will be no dependency conflicts.

SbloodyS: These two dependency only using in test scope. So it doesn't package into dist just like mysql.

streampark-common/src/main/java/org/apache/streampark/common/sql/ClasspathSqlScriptParser.java
```diff
@@ -0,0 +1,147 @@
...
+import java.io.LineNumberReader;
+import java.io.Reader;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+@Slf4j
+public class ClasspathSqlScriptParser implements SqlScriptParser {
```
wolfboys, line 38: I don't know why this class is added, Do we need to execute sql manually?

streampark-registry/streampark-registry-plugins/streampark-registry-it/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.streampark</groupId>
+        <artifactId>streampark-registry-plugins</artifactId>
```
wolfboys, line 23: How about streampark-registry-provider? I don't the name like xxx-plugins

streampark-common/src/main/java/org/apache/streampark/common/sql/ClasspathSqlScriptParser.java
SbloodyS: Because the registry is a separate function, it should be separate from the core function to ensure that the core function is not affected. So it will use a separate database connection pool and a separate thread pool just like `org.apache.streampark.common.thread.BaseDaemonThread`

streampark-registry/streampark-registry-plugins/streampark-registry-it/pom.xml
SbloodyS: Usually in java, provider stands for interface, which is provided to other implementations. But `streampark-registry-plugins` module is not an interface but an implementations.

wolfboys:
Thanks for your contribution. There are some minor issues:

1.  streampark needs to support HA, which is a unique feature of the streampark-console module. As per our architectural design, this feature will depend on the registry module. The registry module is  only used by the console. So, The registry module should not be placed under the root module; instead, it should be a submodule of the console. 

2. I don't think it's necessary to manually execute JDBC in the registry module because the console platform already has an ORM framework. Also, the registry module will be packaged into the console module, so there's no need to introduce a complex design to make the project look clearer.

image

streampark-registry/streampark-registry-plugins/streampark-registry-it/pom.xml
wolfboys: > Usually in java, provider stands for interface, which is provided to other implementations. But `streampark-registry-plugins` module is not an interface but an implementations.

Plugins may create ambiguity, and plugin modules will be introduced later. To avoid confusion, it is not recommended to use xx-plugin.

SbloodyS:
As no consensus can be reached at the moment, temporarily closing this PR.

caicancai:
> Thanks for your contribution. There are some minor issues:
> 
> 1. streampark needs to support HA, which is a unique feature of the streampark-console module. As per our architectural design, this feature will depend on the registry module. The registry module is  only used by the console. So, The registry module should not be placed under the root module; instead, it should be a submodule of the console.
> 2. I don't think it's necessary to manually execute JDBC in the registry module because the console platform already has an ORM framework. Also, the registry module will be packaged into the console module, so there's no need to introduce a complex design to make the project look clearer.
> 
> image

I have a different opinion. We take this opportunity to reconstruct the console module. At present, the coupling of the console module is very exaggerated in my opinion, which is not conducive to the subsequent development of the project.

wolfboys:
> > Thanks for your contribution. There are some minor issues:
> > 
> > 1. streampark needs to support HA, which is a unique feature of the streampark-console module. As per our architectural design, this feature will depend on the registry module. The registry module is  only used by the console. So, The registry module should not be placed under the root module; instead, it should be a submodule of the console.
> > 2. I don't think it's necessary to manually execute JDBC in the registry module because the console platform already has an ORM framework. Also, the registry module will be packaged into the console module, so there's no need to introduce a complex design to make the project look clearer.
> > 
> > image
> 
> I have a different opinion. We take this opportunity to reconstruct the console module. At present, the coupling of the console module is very exaggerated in my opinion, which is not conducive to the subsequent development of the project.

I have never been against splitting the console module or even refactoring the console module to make the project better. Still, these are all internal to the console module, and new modules should not be placed under the root module. You can look at the streampark-flink module, which has many submodules.

image

PR #3965, opened 2024-08-14
Fix the issue where failure to start in k8s mode cannot be stopped.

lintingbin:

The existing k8s mode will change the status to FAILED after startup failure, which will cause the flink job to keep restarting on k8s, and there is no button on the front end of streampark to stop the flink job. This submission is to allow the flink job to be stopped via the abort button on the front end of streampark when the startup fails.

## Brief change log

1. Remove the "K8S_INITIALIZING" state specific to k8s mode.
2. When the rest interface of the flink job is not accessible and the job's deployment exists, the job should enter the Starting state instead of the failed state.
3. Change the abort implementation of k8s mode to delete the deployment.

## Verifying this change


wolfboys (approved):
LGTM

ysn2233:
Why this pr not merged into dev branch?

PR #3967, opened 2024-08-15
[Improve] add streampark task id in external link

ShawHee:
When adding an external link, I found that the parameters only contained information about the remote task, but not the task ID of Streampark itself, which was inconvenient to use. Therefore, I added the ID to the external link.

## Brief change log

add ID in org/apache/streampark/console/core/enums/PlaceholderTypeEnum.java

## Verifying this change

*(or)*

*(or)*


wolfboys (approved):
LGTM

Issue #3983, opened 2024-08-22
[Improvement] Job dependencies limit the default number of uploaded jars


jianjun159:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-streampark/issues?q=is%3Aissue+label%3A%22improvement%22) and found no similar issues.

### Java Version

jdk8

### Scala Version

2.12.x

### StreamPark Version

dev

### Flink Version

flink 1.18.1

### deploy mode

kubernetes-application

### What happened

When I need to rely on external jars, because there are so many jars on the platform that I can't get any more than 25 jars, I want to put this limit in the setttings configuration and let the user decide how many dependencies to display or limit the number

### Screenshots

![image](https://github.com/user-attachments/assets/bfb7c9a8-f6dd-480e-a581-51084f0a5e0e)

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!(您是否要贡献这个PR?)

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wolfboys:
done

PR #3988, opened 2024-08-22
[ISSUES-3983] Job dependencies limit the default number of uploaded jars

jianjun159:

Issue Number: close #xxx 

## Brief change log

The number limit read from the configuration is supported when the job dependent jar is fetched, and the default value is increased to keep the original default value in order to prevent system exceptions

#3983 

 no


streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/SettingServiceImpl.java
```diff
@@ -62,6 +63,8 @@ public class SettingServiceImpl extends ServiceImpl<SettingMapper, Setting>
 
     private final Setting emptySetting = new Setting();
 
+    private static final int DEFAULT_STREAMPARK_UPLOADS_RECORD_LIMIT = 25;
```
wolfboys, line 66: Can the System-setting page (front-end) change this value?

wolfboys:
I don't recommend setting this value in the settings as it complicates usage. The most straightforward approach is to remove this limit, which will not cause other problems.

PR #3989, opened 2024-08-22
[improve] job on yarn state patch.

Mrart:

Issue Number: close #xxx 

## Brief change log

job on yarn state patch.upate log jobmanager url

## Verifying this change

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkAppHttpWatcher.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkAppHttpWatcher.java
@@ -33,6 +33,7 @@ import org.apache.streampark.console.core.metrics.flink.CheckPoints;
 import org.apache.streampark.console.core.metrics.flink.JobsOverview;
 import org.apache.streampark.console.core.metrics.flink.Overview;
 import org.apache.streampark.console.core.metrics.yarn.YarnAppInfo;
+import org.apache.streampark.console.core.service.ApplicationLogService;
 import org.apache.streampark.console.core.service.ApplicationService;
 import org.apache.streampark.console.core.service.FlinkClusterService;
 import org.apache.streampark.console.core.service.SavepointService;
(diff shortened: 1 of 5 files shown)


streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/task/FlinkAppHttpWatcher.java
```diff
@@ -575,6 +578,7 @@ and the status is not obtained this time (flink rest server is closed),
           if (trackingUrl != null && !trackingUrl.equals(application.getJobManagerUrl())) {
             application.setJobManagerUrl(trackingUrl);
             applicationService.updateJobManagerUrl(application.getId(), trackingUrl);
+            applicationLogService.updateJobManagerUrl(application.getId(), trackingUrl);
```
wolfboys: Thanks for your contribution. I think the job history record should not be modified. It is just an objective record of the job. This job instance may have been stopped, so it is unreasonable to modify it.

Mrart: 1. We only need to update the jobmanager url while the job is still running, not killed and cancelled states
2. The job history interface is displayed to use the jobmanagerurl, if you do not update this place to show link errors.

wolfboys: > 1. We only need to update the jobmanager url while the job is still running, not killed and cancelled states
> 2. The job history interface is displayed to use the jobmanagerurl, if you do not update this place to show link errors.

In the t_flink_log table, an application (appid) may be inserted multiple times. Each operation (app build, app start, app stop, app release) will record information about this app. This PR will modify all records of the app, which will result in logical errors.

wolfboys: oh, It is modified by the yarn id. Then there is no problem. Thanks.

wolfboys:
There are some checkstyle issues here. Can you fix it? Thanks

wolfboys (approved):
LGTM

PR #3998, opened 2024-08-27
Fix the "java.nio.file.NoSuchFileException: /root/.kube/config" exception by not providing the Kube Conf File parameter when starting the k8s session cluster.

lintingbin:
## Brief change log

Remove the default value of kubernetes.config.file when starting the k8s session cluster (similar to k8s application). When the default value of kubernetes.config.file is not provided, StreamPark will use the service account permissions of the pod it resides in to start the Flink cluster, without the need for the /root/.kube/config file.

## Verifying this change

*(or)*

*(or)*


Mrart:
You mean just get the kubeconfig environment variable at startup?

lintingbin:
> You mean just get the kubeconfig environment variable at startup?
image
When starting, you can specify the "Kube Conf File" or leave it unspecified by setting "Kube Conf File" to null. If the Flink session and StreamPark are in the same k8s cluster, the "Kube Conf File" can be set to null. In this case, the Flink session will be started using the service account permissions of the StreamPark pod (this is how Flink application is currently started). Currently, the code sets the default value of "kubeConf" to "~/.kube/config", but the kubeConf is not mandatory. A Flink session can be started without the kubeConf, just like starting a Flink application without providing a kubeConf.

wolfboys (approved):

wolfboys (approved):

PR #4016, opened 2024-09-01
[Feature-3934] Add registy module for streampark

### LINKED ISSUE #3934: [Feature] Add registy module for streampark

HxpSerein:

Issue Number: close https://github.com/apache/incubator-streampark/issues/3934

## Brief change log
Add streampark-registy module

## Verifying this change

- NetworkUtilsTestCase
- UUIDUtilsTestCase
- RegistryTestCase
- JdbcRegistryTestCase
- JdbcRegistryThreadFactoryTest
- LockUtilsTest
- MysqlJdbcRegistryTestCase
- PostgresqlJdbcRegistryTestCase

 - Dependencies (does it add or upgrade a dependency): yes


streampark-common/src/main/java/org/apache/streampark/common/enums/ServerStatus.java
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.streampark.common.enums;
+
+public enum ServerStatus {
```
wolfboys: streampark-common is a global module. Avoid putting console layer logic into streampark-common.

streampark-common/src/main/java/org/apache/streampark/common/lifecycle/ServerLifeCycleException.java
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.streampark.common.lifecycle;
+
+public class ServerLifeCycleException extends Exception {
```
wolfboys: This logic should not be placed in the common module; it belongs in the console-service layer.

streampark-common/src/main/java/org/apache/streampark/common/enums/ServerStatus.java
HxpSerein: I have moved the relevant classes from `streampark-common` to `streampark-console-registry`.

streampark-common/src/main/java/org/apache/streampark/common/lifecycle/ServerLifeCycleException.java
HxpSerein: Moved.

wolfboys (approved):
LGTM

ruanwenjun:
@wolfboys @HxpSerein Hi, I find most of the code under `streampark-console-registry` is copy from DS, you should announce this.

HxpSerein:
@ruanwenjun Thank you for reporting this issue. After review, we confirmed that the issue does exist. This code is only in the dev branch. We will list the code in the LICENSE. This is only a temporary solution and we will rewrite this part of the code later. Thank you again for your feedback.

wolfboys:
hi @ruanwenjun @HxpSerein @SbloodyS:

This pull request is a fork of [another PR](https://github.com/apache/incubator-streampark/pull/3948), After review, we discovered that the code in the `streampark-console-registry` module was copied from [Apache DolphinScheduler](https://github.com/apache/dolphinscheduler/tree/dev/dolphinscheduler-registry) and not announce this, This violates the Apache License specifications. Therefore, we have submitted a [new PR](https://github.com/apache/incubator-streampark/pull/4122) to reverse the changes made in the previous request. We are very sorry for any inconvenience this may cause.

PR #4029, opened 2024-09-04
Fix the issue where the HA files are not cleaned up when the Flink job stops, causing two identical jobs to start the next time a new job is launched.

lintingbin:
…b stops, causing two identical jobs to start the next time a new job is launched.

## Brief change log

Fix the issue where the HA files are not cleaned up when the Flink job stops, causing two identical jobs to start the next time a new job is launched.

## Verifying this change

*(or)*

*(or)*


wolfboys (approved):
LGTM

PR #4030, opened 2024-09-04
Fix the issue of state update failure.

lintingbin:
## Brief change log

The subscribeMetricsChange method reads the current state from the database and then updates it back to the database, which may cause the subscribeJobStatusChange update to fail. Therefore, the persistMetrics function has been modified so that the state field is not updated every time; it is only updated when the state field needs modification.

## Verifying this change

*(or)*

*(or)*


wolfboys (approved):
LGTM

PR #4040, opened 2024-09-06
[Feature-3960] Add task distribution for streampark

### LINKED ISSUE #3960: [Feature] Add task distribution for streampark
### Search before asking

### Description

Distribute the task to the specified server, using the consistent hashing algorithm by default.

The previous design suggested separating task startup and monitoring. While the task could be started on any server, monitoring needed to be distributed to maintain load balancing. However, at the current stage, there is a dependency between task startup and monitoring, and separating the two would require additional design efforts. Additionally, during fault recovery, task startup would also need to be reallocated, adding further complexity. To minimize changes in the first stage, task startup, shutdown, and monitoring will all be performed on the same specified server. The following is the specific design:

<div align="center">
  
</div>

- **Provide the `ConsistentHash` utility class**: Support adding and deleting servers, and hash tasks to the specified server based on the task ID.
- **Implement a unified `FlinkTaskService` interface**: Include functionality for obtaining a list of monitored tasks, starting/stopping tasks, and redistributing tasks.
- **Create a new table `t_flink_task` in the database for message communication**: Use the producer-consumer model, writing a record to the table when a task starts/stops. FlinkTaskService will poll the table to obtain task records and use ConsistentHash to determine if the task belongs to it. If not, it takes no action; if it does, it performs the relevant operations.

### Function

The SP uses task distribution to perform the following three operations:

- Use the producer-consumer model to start/stop tasks.
- When there is no server addition or deletion, ConsistentHash ensures the stability of the hash result, with each task being assigned to only one specified server.
- When a server is added or removed, the registry center triggers FlinkTaskService to reallocate tasks.

### How to use
SP automatically manages the distribution of tasks without the user's awareness.

### Module

**ConsistentHash:** Add a utility class for the task distribution algorithm.

```java 
public class ConsistentHash<T> {

    // the number of virtual nodes for each server
    private final int numberOfReplicas = 2 << 16;

    // the hash ring of servers
    private final SortedMap<Long, T> circle = new TreeMap<>();

    /**
     * Initialize the ConsistentHash with a collection of servers.
     */
    public ConsistentHash(Collection<T> servers) { servers.forEach(this::add); }

    /**
     * Add the virtual nodes of the server to the hash ring.
     */
    public void add(T server) {
        for (int i = 0; i < numberOfReplicas; i++) {
            circle.put(Murmur3Hash.hash64(server.toString() + i), server);
        }
    }

    /**
     * Remove the virtual nodes of the server from the hash ring.
     */
    public void remove(T server) {
        for (int i = 0; i < numberOfReplicas; i++) {
            circle.remove(Murmur3Hash.hash64(server.toString() + i));
        }
    }

    /**
     * Get the server that the key belongs to from the hash ring.
     */
    public T get(Object key) {
        if (circle.isEmpty()) {
            return null;
        }
        long hash = Murmur3Hash.hash64(key.toString());
        if (!circle.containsKey(hash)) {
            SortedMap<Long, T> tailMap = circle.tailMap(hash);
            hash = tailMap.isEmpty() ? circle.firstKey() : tailMap.firstKey();
        }
        return circle.get(hash);
    }
}
```
**FlinkTaskService:** This module includes the relevant interfaces involved in task distribution.

```java
/**
 * FlinkTaskService is the interface for managing tasks.
 */
public interface FlinkTaskService extends IService<FlinkTask> {

    /**
     * Add the current console server itself to the consistent hash ring.
     * @param serverName String
     */
    void init(String serverName);

    /**
     * This interface is responsible for polling the database to retrieve task records and execute the corresponding operations.
     * @param flinkTask FlinkTask
     */
    void executeFlinkTask(FlinkTask flinkTask) throws Exception;

    /**
     * Through this interface, the watcher obtains the list of tasks that need to be monitored.
     * @param applications List<Application>
     * @return List<Application> List of tasks that need to be monitored
     */
    List<Application> getMonitoredTaskList(List<Application> applications);

    /**
     * This interface handles task redistribution when server nodes are added.
     * @param server String
     */
    void addServerRedistribute(String server);

    /**
     * This interface handles task redistribution when server nodes are removed.
     * @param server String
     */
    void removeServerRedistribute(String server);

    /**
     * Determine whether the task is processed locally
     *
     * @param appId Long
     * @return boolean
     */
    public boolean isLocalProcessing(Long appId);

    /**
     * Save flink task
     *
     * @param appParam  Application
     * @param autoStart boolean
     * @param action It may be one of the following values: START, RESTART, REVOKE, CANCEL, ABORT
     */
    public void saveFlinkTask(Application appParam, boolean autoStart, FlinkTaskEnum action);
}
```

### Usage Scenario

_No response_

### Related issues

_No response_

### Code of Conduct

HxpSerein:

Issue Number: close https://github.com/apache/incubator-streampark/issues/3960

## Brief change log
Add task distribution for streampark

## Verifying this change

- ConsistentHashTest
- FlinkTaskServiceTest


Mrart:
Thanks for contributing, I'm leaving a few comments, 
1) We only support flink yarn jobs? we need supported k8s and spark jobs.
2) We can have more test cases to protect the stability of the program. So far we have only tested consistent hashes,.

HxpSerein:
Thank you, @Mrart, for your review.

I fully agree with your suggestion to add more test cases to enhance the program's stability. So far, we've only tested consistent hashing, and additional tests for other scenarios will be helpful moving forward.

> We can have more test cases to protect the stability of the program. So far we have only tested consistent hashes.

This PR modifies the task distribution logic in ApplicationActionServiceImpl. For single-node scenarios, the current logic will still be executed locally. In multi-node HA scenarios, since Streampark currently doesn’t support multi-node integration testing (IT), and adding IT tests in the short term is challenging, I plan to add unit tests (UT) for logic correctness. Once the resource center is integrated, we can proceed with multi-node cluster deployment testing.

> We only support flink yarn jobs? we need supported k8s and spark jobs.

Since the resource center is not yet developed and HA is incomplete, the current testing support is somewhat limited. To ensure code stability, I propose to temporarily support only Flink tasks, with support for Kubernetes and Spark jobs added once the resource center is completed. Do you have any suggestions on this approach? @Mrart @wolfboys

Mrart:
> Since the resource center is not yet developed and HA is incomplete, the current testing support is somewhat limited. To ensure code stability, I propose to temporarily support only Flink tasks, with support for Kubernetes and Spark jobs added once the resource center is completed. Do you have any suggestions on this approach? @Mrart @wolfboys

I agree it.

Mrart:
Another suggestion is that we merge our commits before we merge them and change the commit to match what we changed

streampark-console/streampark-console-service/src/main/assembly/script/schema/mysql-schema.sql
```diff
@@ -240,6 +240,19 @@ create table `t_flink_sql` (
 ) engine=innodb auto_increment=100000 default charset=utf8mb4 collate=utf8mb4_general_ci;
 
 
+-- ----------------------------
+-- Table structure for t_ha_task
+-- ----------------------------
+drop table if exists `t_ha_task`;
+create table `t_ha_task` (
```
wolfboys: I suggest the table name rename to  `t_console_ha_task`

streampark-console/streampark-console-service/src/main/assembly/script/schema/mysql-schema.sql
```diff
@@ -240,6 +240,19 @@ create table `t_flink_sql` (
 ) engine=innodb auto_increment=100000 default charset=utf8mb4 collate=utf8mb4_general_ci;
 
 
+-- ----------------------------
+-- Table structure for t_distribution_task
+-- ----------------------------
+drop table if exists `t_distribution_task`;
```
wolfboys: Would it be better to rename table "t_distributed_task"?

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/DistributionTaskServiceImpl.java
```diff
@@ -0,0 +1,239 @@
...
+@Service
+@Transactional(propagation = Propagation.SUPPORTS, readOnly = true, rollbackFor = Exception.class)
+public class DistributionTaskServiceImpl extends ServiceImpl<DistributionTaskMapper, DistributionTask>
+    implements
+        DistributionTaskService {
+
+    /**
+     * Server name
+     */
+    private String serverName;
```
wolfboys: It is recommended to rename to "serverId"

streampark-console/streampark-console-service/src/main/resources/mapper/core/DistributionTaskMapper.xml
```diff
@@ -0,0 +1,28 @@
...
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+<!DOCTYPE mapper
+        PUBLIC "-//mybatis.org//DTD Mapper 3.0//EN"
+        "http://mybatis.org/dtd/mybatis-3-mapper.dtd">
+<mapper namespace="org.apache.streampark.console.core.mapper.DistributionTaskMapper">
+    <resultMap id="BaseResultMap" type="org.apache.streampark.console.core.entity.DistributionTask">
```
wolfboys: We don't need to define this BaseResultMap because it has never been used.

HxpSerein: Resolved.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/DistributionTaskServiceImpl.java
HxpSerein: Resolved.

streampark-console/streampark-console-service/src/main/assembly/script/schema/mysql-schema.sql
HxpSerein: Resolved.

wolfboys (approved):
LGTM

PR #4127, opened 2024-10-29
[Feature-3934] Add zookeeper registry for streampark

### LINKED ISSUE #3934: [Feature] Add registy module for streampark

HxpSerein:

Issue Number: close https://github.com/apache/incubator-streampark/issues/3934

## Brief change log

Add zookeeper registry for streampark

## Verifying this change

RegistryServiceTest

 - Dependencies (does it add or upgrade a dependency): yes

### DIFF
--- a/streampark-console/streampark-console-service/src/main/assembly/conf/config.yaml
+++ b/streampark-console/streampark-console-service/src/main/assembly/conf/config.yaml
@@ -106,6 +106,10 @@ registry:
   heartbeat-refresh-interval: 1s
   session-timeout: 3s
 
+zookeeper:
+  # zookeeper address
+  address: localhost:2181
+
 network:
   # network interface preferred like eth0, default: empty
   preferred-interface: ""
--- a/streampark-console/streampark-console-service/src/main/assembly/conf/config.yaml
+++ b/streampark-console/streampark-console-service/src/main/assembly/conf/config.yaml
@@ -100,15 +100,13 @@ sso:
         # Optional, change by authentication client
         # Please replace and fill in your client config below when enabled SSO
 
-registry:
-  # default using jdbc as registry
-  type: jdbc
-  heartbeat-refresh-interval: 1s
-  session-timeout: 3s
-
-zookeeper:
-  # zookeeper address
-  address: localhost:2181
+high-availability:
+  enable: false # true
+  # The list of ZooKeeper quorum peers that coordinate the high-availability
+  # setup. This must be a list of the form:
+  # "host1:clientPort,host2:clientPort,..." (default clientPort: 2181)
+  #
+  zookeeper.quorum: 192.168.100.128:2181,192.168.100.129:2181
 
 network:
   # network interface preferred like eth0, default: empty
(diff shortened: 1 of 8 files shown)


streampark-console/streampark-console-service/src/main/assembly/conf/config.yaml
```diff
@@ -106,6 +106,10 @@ registry:
   heartbeat-refresh-interval: 1s
   session-timeout: 3s
 
+zookeeper:
```
wolfboys: Directly configuring Zookeeper will confuse users, and they will need to learn what it does. It is recommended to change to the following:
```

high-availability:
    enable: false # true
    # The list of ZooKeeper quorum peers that coordinate the high-availability
    # setup. This must be a list of the form:
    # "host1:clientPort,host2:clientPort,..." (default clientPort: 2181)
    #
    zookeeper.quorum: 192.168.100.128:2181,192.168.100.129:2181
```

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/StreamParkConsoleBootstrap.java
```diff
@@ -48,11 +52,23 @@
...
     public static void main(String[] args) throws Exception {
         new SpringApplicationBuilder()
             .properties(SpringProperties.get())
             .sources(StreamParkConsoleBootstrap.class)
             .run(args);
     }
 
+    @PostConstruct
+    public void init() {
+        registryService.start();
```
wolfboys: We need to decide whether to call this method based on the configuration value "high-availability.enable."

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/RegistryService.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package org.apache.streampark.console.core.service;
+
+public interface RegistryService {
+
+    /**
+     * Start the registry service.
+     */
+    void start();
```
wolfboys: It is recommended to rename the method name to `register`

streampark-console/streampark-console-service/src/main/assembly/conf/config.yaml
HxpSerein: changed

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/StreamParkConsoleBootstrap.java
HxpSerein: changed

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/RegistryService.java
HxpSerein: ok

wolfboys (approved):
LGTM

PR #4149, opened 2024-12-19
[Bug] Fix the issue where metrics do not update upon restart.

Jam804:
[Bug] Fix the issue where metrics do not update upon restart.


wolfboys (approved):
LGTM

PR #4254, opened 2025-06-08
[Improve] Optimize the job state when alerting

Gianzie:

Job is "FINISHED", but alert is "CANCELED"
By the way, when the job is "FINISHED", it seems that there is no need for an alert. What do you think?@wolfboys

## Verifying this change

*(or)*

*(or)*


wolfboys (approved):
LGTM

PR #4273, opened 2025-08-01
[Improve] Improve login authentication

wolfboys:

Issue Number: close #xxx 

## Verifying this change

*(or)*

*(or)*


streampark-console/streampark-console-service/src/main/assembly/bin/streampark.sh
```diff
@@ -517,6 +519,13 @@ stop() {
   fi
 }
 
+jwt_secret() {
+  local secret=`$JAVACMD -cp "$APP_LIB/*" $BASH_UTIL --jwt_secret`
+  if [[ -n "$secret" ]]; then
+    echo "JWT_SECRET: $secret"
```
omkarparth: Hi, I am the security researcher who reported the security issues which are being fixed as part of this PR. 
On the reviewing the changes, it looks like the encryption key is being logged in service logs. 
This is not recommended as it will create a footprint of the encryption key in service logs. And anyone (including internal threat actors) who have access to the service logs would be able to access the encryption key and decrypt the JWT tokens.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/util/BashJavaUtils.java
```diff
@@ -46,6 +49,12 @@ public static void main(String[] args) throws IOException {
         String[] actionArgs = Arrays.copyOfRange(args, 1, args.length);
 
         switch (action) {
+            case "--jwt_secret":
+                String secret = getFirstOrCreateJWTSecret();
+                if (secret != null) {
+                    System.out.println(secret);
```
omkarparth: Same issue as above
This creates another footprint of the JWT secret on console.

streampark-console/streampark-console-service/src/main/assembly/bin/streampark.sh
wolfboys: Thanks for the review, updated.

streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/base/util/BashJavaUtils.java
wolfboys: done

GOODBOY008 (approved):
LGTM

PR #3493, opened 2024-01-17
[Feature] Trigger method add savepoint save

gwy1995:
## Contribution Checklist

  - If this is your first time, please read our contributor guidelines: [Submit Code](https://streampark.apache.org/community/submit_guide/submit_code).

  - Make sure that the pull request corresponds to a [GITHUB issue](https://github.com/apache/incubator-streampark/issues).

  - Name the pull request in the form "[Feature] Title of the pull request", where *Feature* can be replaced by `Hotfix`, `Bug`, etc.

  - Fill out the template below to describe the changes contributed by the pull request. That will give reviewers the context they need to do the review.

  - If the PR is unfinished, add `[WIP]` in your PR title, e.g., `[WIP][Feature] Title of the pull request`.

## Brief change log

- *Add savepoint record to t_savepoint when trigger savepoint*

## Verifying this change

This change is a trivial rework / code cleanup without any test coverage.

## Does this pull request potentially affect one of the following parts
 - Dependencies (does it add or upgrade a dependency): no

PR #4280, opened 2025-08-30
[Fix] Manually trigger savepoint and save after success

CuteShaos:
I noticed that this issue has been fixed in previous versions, but I found that 2.1.6 still has this problem. Through Git history, I found that this part of the code was deleted later. What is the reason for this? Although it is difficult to use this feature frequently, when manually triggering savepoint, I found that there are operation logs but no savepoint path recorded, so I cannot use this savepoint to start the Flink task

Issue Number: close #3493 

## Verifying this change


wolfboys (approved):
LGTM

niumy0701:
I have a question. After the savepoint is successful, the disk has already been stored here， and there are no issues with the 2.1.6 verification  @CuteShaos @wolfboys 
image

Issue #4072, opened 2024-09-19
[Bug] email check method is error


sanwen1221:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-streampark/issues?q=is%3Aissue+label%3A%22bug%22) and found no similar issues.

### Java Version

1.8

### Scala Version

2.12.x

### StreamPark Version

2.1.4

### Flink Version

1.17.1

### deploy mode

None

### What happened

到底是选tls还是ssl应该让用户去选择，现在我无法使用tls通过验证
![image](https://github.com/user-attachments/assets/766f5081-409d-4d81-adb2-235975a4b104)
我觉得这样更合适

### Error Exception

```log
Could not connect to SMTP host: smtp.126.com, port: 994, response: -1
```

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!(您是否要贡献这个PR?)

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wolfboys:
Thanks for your feedback. We are closing this issue because it is outdated.

PR #4363, opened 2026-06-22
[ISSUE-4072][Bug] Use SSL when checking alert email

hutiefang76:
## What changed

This fixes the alert email connection check for SSL SMTP settings.

The actual email sending path uses SSL-on-connect when `SenderEmail.ssl` is enabled, but `SettingServiceImpl.checkEmail` was enabling STARTTLS instead. That makes the check fail for SSL SMTP servers such as the case reported in #4072.

This patch extracts the JavaMail property building and makes `checkEmail` set `mail.smtp.ssl.enable=true` when SSL is selected. Plain SMTP keeps the existing behavior.

## Tests

```bash
JAVA_HOME=$(/usr/libexec/java_home -v 17) ./mvnw -s /tmp/codex-empty-maven-settings.xml \
  -pl streampark-console/streampark-console-service -am \
  -DskipITs -DskipRat -DskipWeb -DfailIfNoTests=false \
  -Dtest=SettingServiceImplTest test
```

Also checked:

```bash
git diff --check
```

I used OpenAI Codex to help inspect the issue and prepare the patch, then reviewed the diff and ran the tests above locally.

### DIFF
--- a/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/SettingServiceImpl.java
+++ b/streampark-console/streampark-console-service/src/main/java/org/apache/streampark/console/core/service/impl/SettingServiceImpl.java
@@ -225,14 +225,7 @@ public class SettingServiceImpl extends ServiceImpl<SettingMapper, Setting>
     @Override
     public ResponseResult checkEmail(SenderEmail senderEmail) {
         ResponseResult result = new ResponseResult();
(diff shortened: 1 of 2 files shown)


hutiefang76:
Closing this PR because the linked issue was closed by the project as outdated on September 3. The branch has also diverged from current dev, so rebasing it would add noise without an active upstream request. Thanks for the review time.

Issue #4408, opened 2026-06-30
[Proposal] Remove Scala from StreamPark — design and migration plan


shangeyao:
## Background

StreamPark was originally built with a Scala-first development model: user applications extend `FlinkStreaming` / `FlinkTable` / `SparkStreaming` / `SparkBatch` traits; Flink multi-version support is implemented via Scala shims behind `FlinkShimsProxy`; and most runtime modules (`streampark-common`, Flink client/K8s/packer, Spark client) are written in Scala.

This creates several long-term costs:

1. **Build complexity**: `scala-maven-plugin`, Scalafmt, Scalastyle, silencer, ScalaTest, and mixed Java/Scala compilation increase CI time and contributor onboarding friction.
2. **Ecosystem alignment**: Apache Flink 2.x removed the Scala DataStream API; continuing Scala-centric APIs fights upstream direction.
3. **Console alignment**: `streampark-console-service` is Java/Spring Boot but depends heavily on Scala-compiled libraries (`HadoopUtils`, `FlinkClient`, `FlinkShimsProxy`, packer pipelines, etc.).
4. **Maintenance multiplier**: Every shims change must be replicated across many Flink version modules, most of which are Scala.

**#4432** removed the user-facing dev framework and connector wrappers. Flink SQL jobs now run through **`SqlClient`** → versioned **`StreamTableContext` / `TableContext`** → **`FlinkSqlExecutor`**. Structural refactors on that path must preserve external behavior.

This issue tracks the **design and execution plan** to remove Scala from the StreamPark codebase and safely refactor the Flink SQL submission stack.

**Implementation PRs:** #4458, #4461–#4467 merged; #4471 in review (Flink post-migration build cleanup)  
**Depends on:** #4409 (JDK 11 baseline ✅), #4432 (dev framework removal ✅)  
**Parent roadmap:** #4410

**Community workflow** (PMC sync): migrate **class-by-class**; each new Java class should ship with tests (or document why not). Sub-task issues: #4445 #4446 #4447 #4448 #4449 #4450 #4451 #4452 #4453 #4470 — link PRs to the matching row below. Opening a separate GitHub issue per class is optional when you want an isolated review thread.

---

## Progress Update (2026-08-03)

[...]

PR #4432, opened 2026-07-14
[Flink] Remove dev framework and connector wrappers

shangeyao:
## Summary

Remove the StreamPark development framework ([framework docs](https://streampark.apache.org/docs/framework/)) and all connector wrapper modules. Console platform runtime (submit/cancel, SQL mode, validation, K8s, packaging) is preserved.

Closes #4431.

## Changes

- Remove `streampark-flink-core` (`FlinkStreaming`, `FlinkTable`, `FlinkStreamTable`, `DataStreamExt`)
- Remove entire `streampark-flink-connector` tree (Kafka, JDBC, ClickHouse, Redis, ES, HBase, HTTP, etc.)
- Remove Spark dev traits (`Spark`, `SparkBatch`, `SparkStreaming`) and `streampark-spark-connector`
- Remove Maven `-Pconnector` profile
- Refactor `Flink SqlClient` to use `FlinkTableInitializer` + `TableContext` / `StreamTableContext` from shims
- Add `SparkSqlExecutor` and refactor `Spark SqlClient` accordingly

## Breaking changes

- User jobs extending StreamPark dev traits must migrate to native Flink/Spark APIs
- User code using `org.apache.streampark.flink.connector.*` or `org.apache.streampark.spark.connector.*` must migrate to official connectors

## Preserved (non-breaking for Console)

- Flink/Spark SQL job submission via `SqlClient`
- SQL validation (`FlinkSqlValidator`, `SparkSqlValidator`)
- Flink client, packer, kubernetes, shims

## Test plan

- [x] `./mvnw -Pfast clean compile -pl streampark-flink/streampark-flink-sqlclient,streampark-spark/streampark-spark-sqlclient,streampark-spark/streampark-spark-core -am -DskipTests` (JDK 8)
- [x] `./mvnw -Pfast clean compile -pl streampark-console/streampark-console-service -am -DskipTests` (JDK 8)
- [ ] Full CI
- [ ] Manual: Console Flink SQL / Spark SQL submit on YARN or K8s

Made with [Cursor](https://cursor.com)

PR #4436, opened 2026-07-19
[Feature][Flink] Define FlinkStreamingJob Java lifecycle base class

och5351:

Implements [#4435](https://github.com/apache/streampark/issues/4435) — Phase 0.1 of the Scala removal plan ([#4408](https://github.com/apache/streampark/issues/4408)).

Introduces a pure-Java `streampark-flink-core` module and defines `FlinkStreamingJob` as the Java abstract base class that replaces the deleted Scala `FlinkStreaming` trait.

## Brief change log

<details>
    <summary>previous code</summary>
  
```scala
package org.apache.streampark.flink.core.scala

class StreamingContext(val parameter: ParameterTool, private val environment: StreamExecutionEnvironment)
  extends StreamExecutionEnvironment(environment.getJavaEnv) {
  def start(): JobExecutionResult = execute()
  override def execute(): JobExecutionResult = {
    val appName = parameter.getAppName(required = true)
    execute(appName)
  }
}

trait FlinkStreaming extends Serializable with Logger {
  implicit final lazy val parameter: ParameterTool = context.parameter
  implicit var context: StreamingContext = _
  var jobExecutionResult: JobExecutionResult = _

  final def main(args: Array[String]): Unit = {
    init(args)
    ready()
    handle()
    jobExecutionResult = context.start()
    destroy()
  }

  private[this] def init(args: Array[String]): Unit = {
    SystemPropertyUtils.setAppHome(KEY_APP_HOME, classOf[FlinkStreaming])
    context = new StreamingContext(FlinkStreamingInitializer.initialize(args, config))
  }

  def ready(): Unit = {}
  def config(env: StreamExecutionEnvironment, parameter: ParameterTool): Unit = {}
  def handle(): Unit
  def destroy(): Unit = {}
}
```

</details>

- Add new Maven module `streampark-flink/streampark-flink-core` (no Scala version suffix, no Scala dependencies)
- Define `FlinkStreamingJob` abstract class with lifecycle: `configure → ready → handle → execute → destroy`
- Register the new module in `streampark-flink/pom.xml`
- Add JUnit 5 tests verifying lifecycle order, class name default, and `ParameterTool` propagation

## Verifying this change

```bash
mvn test -pl streampark-flink/streampark-flink-core
```

Tests run: 3, Failures: 0, Errors: 0

## Does this pull request potentially affect one of the following parts?

- [x] The public API
- [ ] The web application
- [ ] The build system

## Documentation

- Relates to issue [#4435](https://github.com/apache/streampark/issues/4435)
- Part of umbrella issue [#4408](https://github.com/apache/streampark/issues/4408)

### DIFF
--- a/streampark-flink/streampark-flink-core/src/main/java/org/apache/streampark/flink/core/FlinkStreamingJob.java
+++ b/streampark-flink/streampark-flink-core/src/main/java/org/apache/streampark/flink/core/FlinkStreamingJob.java
@@ -1,0 +1,60 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.streampark.flink.core;
+
+import org.apache.flink.api.common.JobExecutionResult;
+import org.apache.flink.api.java.utils.ParameterTool;
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+
+public abstract class FlinkStreamingJob {
+
+    protected StreamExecutionEnvironment env;
+    protected ParameterTool parameter;
+    protected JobExecutionResult jobExecutionResult;
+
+    public final void run(String[] args) throws FlinkJobException {
+        try {
+            init(args);
+            ready();
+            handle();
+            jobExecutionResult =
+                env.execute(parameter.get("app.name", getClass().getSimpleName()));
+            destroy();
+        } catch (Exception e) {
+            throw new FlinkJobException("Failed to run Flink job: " + getClass().getSimpleName(), e);
+        }
+    }
+
+    private void init(String[] args) {
+        this.parameter = ParameterTool.fromArgs(args);
+        this.env = StreamExecutionEnvironment.getExecutionEnvironment();
+        configure(env, parameter);
+        env.getConfig().setGlobalJobParameters(parameter);
+    }
+
+    protected void configure(StreamExecutionEnvironment env, ParameterTool parameter) {
+    }
+
+    protected void ready() {
+    }
+
+    protected abstract void handle() throws FlinkJobException;
+
+    protected void destroy() {
+    }
+}
--- a/streampark-flink/streampark-flink-core/pom.xml
+++ b/streampark-flink/streampark-flink-core/pom.xml
@@ -20,66 +20,48 @@
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>org.apache.streampark</groupId>
-        <artifactId>streampark</artifactId>
+        <artifactId>streampark-flink</artifactId>
         <version>${revision}</version>
     </parent>
 
-    <artifactId>streampark-flink</artifactId>
-    <packaging>pom</packaging>
-    <name>StreamPark : Flink Parent</name>
-
-    <modules>
-        <module>streampark-flink-core</module>
-        <module>streampark-flink-shims</module>
-        <module>streampark-flink-sqlclient</module>
-        <module>streampark-flink-udf</module>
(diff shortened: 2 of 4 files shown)


streampark-flink/streampark-flink-core/src/main/java/org/apache/streampark/flink/core/FlinkStreamingJob.java
```diff
@@ -0,0 +1,60 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.streampark.flink.core;
+
+import org.apache.flink.api.common.JobExecutionResult;
+import org.apache.flink.api.java.utils.ParameterTool;
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+
+public abstract class FlinkStreamingJob {
```
wolfboys, line 24: We greatly appreciate your contribution. I have observed that the current Java implementation differs considerably from the Scala version, and that several core classes (FlinkStreamTable, FlinkTable) are not yet updated. The recommendation is to keep the original Scala module's logic unchanged and only reimplement it in Java, which would reduce the risk to a minimum. Subsequent iterations and modifications can be made thereafter.

streampark-flink/streampark-flink-core/.flattened-pom.xml
wolfboys, line 1: Could you please check if the .flattened-pom.xml file is needs to be committed?

streampark-flink/streampark-flink-core/src/main/java/org/apache/streampark/flink/core/FlinkStreamingJob.java
och5351: Thank you for the review @wolfboys !

1. Regarding FlinkStreamTable and FlinkTable — this PR intentionally covers only Phase 0.1 of the #4408 roadmap, limiting the scope to FlinkStreamingJob. The remaining classes are planned for subsequent phases. I apologize if that context wasn't clear enough in the PR description. If my approach is off track, I would really appreciate your feedback.

2. Regarding the difference from the Scala version — the lifecycle structure (init → ready → handle → execute → destroy) follows the original. The differences are intentional: since FlinkStreamingInitializer and StreamingContext are themselves Scala components targeted for removal in later phases, the Java version uses Flink's standard APIs directly instead of depending on them. This was a deliberate decision to avoid StreamPark's own wrapper classes, but I'd love your thoughts on whether this direction is the right one.

streampark-flink/streampark-flink-core/.flattened-pom.xml
och5351: Thank you for your notice :)

shangeyao:
Thanks for picking up #4408 Phase 0.1 — the scoped approach and lifecycle tests look good.

Before merge, I think we need closer parity with the old `FlinkStreaming` platform contract:

1. **Init path** — Wire init through `FlinkStreamingInitializer` (or an equivalent Java port): `--conf`, `setAppHome`, merged `ParameterTool` / `FlinkConfiguration`. The current `ParameterTool.fromArgs()` + bare `StreamExecutionEnvironment.getExecutionEnvironment()` won't work for Console-submitted jobs.
2. **App name** — Use StreamPark app-name resolution (`getAppName` semantics: `app.name`, `pipeline.name`, deflate), not only raw `parameter.get("app.name", ...)`. The existing test doesn't verify the name passed to `execute()`.
3. **`main()` entry** — Console launches user `mainClass`; the old trait provided `main()`. Please document or provide a standard entry pattern for `FlinkStreamingJob`.
4. **Context vs #4432** — Clarify in the PR description that this reintroduces `streampark-flink-core` as the **Java replacement** after #4432, per #4408 — not a rollback.

Non-blocking: consider `finally` for `destroy()`, align package name with #4408 (`core.java`?), add `streampark-common` (provided).

Happy to re-review once the blocking items are addressed. +1 on phased follow-ups for `FlinkTable` / `FlinkStreamTable`.

wolfboys:
Thank you for your valuable contribution.
After careful evaluation, we found that the current PR introduces significant changes compared with the existing implementation and may impact the original design and migration approach. To ensure a minimal, stable, and manageable migration process, the PMC has discussed and reviewed the proposal, and we have redefined the roadmap for removing Scala.

Currently, we plan to proceed with Stage 1 — Flink Core: [4445](https://github.com/apache/streampark/issues/4445) ,  [4446](https://github.com/apache/streampark/issues/4446) ,  [4447](https://github.com/apache/streampark/issues/4447) , [4448](https://github.com/apache/streampark/issues/4448) , migrating modules incrementally according to the proposed execution plan(Principle: no logic changes; migrate modules to Java with readable, idiomatic code). We kindly suggest following the steps outlined in the issue: [flink core](https://github.com/apache/streampark/issues/4408)

Thanks again for your effort and contribution. We look forward to continuing the collaboration on this migration.

och5351:
@wolfboys 

Thanks for providing the roadmap. I'll begin with issue #4445 and work on making meaningful contributions to the project.

PR #4441, opened 2026-07-21
[Feature][Flink] Define FlinkTableJob and FlinkStreamTableJob Java base classes

ocb3916:

Implements Phase 0.2 of the Scala removal plan (#4408).

Defines `FlinkTableJob` and `FlinkStreamTableJob` as the Java abstract base classes that will
eventually replace the Scala `FlinkTableTrait` and `FlinkStreamTableTrait` traits — for Table API
(batch) jobs and jobs that mix the DataStream API with the Table API, respectively. Both classes
live in the existing `streampark-flink-shims-base` module and follow the same
`ready → handle → destroy` lifecycle established by `FlinkStreamingJob` in sub-issue 0.1, so all
three "Job" base classes share a consistent shape.

## Brief change log

- Add `FlinkTableJob` abstract class with lifecycle: `ready → handle → destroy`, exposing the
  underlying `TableEnvironment` via `getTableEnv()` (composition) instead of extending / mimicking
  the interface like the legacy trait did
- Add `FlinkStreamTableJob` abstract class with the same lifecycle, exposing both
  `StreamExecutionEnvironment` (`getEnv()`) and `StreamTableEnvironment` (`getTableEnv()`) via
  composition
- Add `sql(String)` / `sql(String, Consumer<String>)` on both classes, bridging to the existing
  Scala `FlinkSqlExecutor.executeSql` so the public API surface stays Scala-free
- Add `markConvertedToDataStream()` on `FlinkStreamTableJob`, replacing the legacy trait's
  automatic `Table → DataStream` conversion detection (no longer possible under composition) with
  an explicit call
- Place both classes under `org.apache.streampark.flink.core.javaapi` rather than the
  issue-specified `org.apache.streampark.flink.core.java` — see note below
- Add JUnit 5 tests for both classes verifying lifecycle order, `getTableEnv()`/`getEnv()`
  composition, SQL execution, and the `markConvertedToDataStream()` flag
- Extract shared lifecycle/logic to fix SonarCloud duplication warnings:
  - `AbstractFlinkJob` (package-private) — common `ready → handle → execute →
    destroy` lifecycle, `sql(...)`, and `app.name` lookup shared by both job classes
  - `FlinkJobSupport` — SQL callback bridging (Java `Consumer` <-> Scala `Function1`)
  - `JobTestParams` — shared `ParameterTool` builder used by both test classes
- Replace `System.out.println` with SLF4J logging in both `execute(...)` overrides

## A note on the package name

The issue body specifies `org.apache.streampark.flink.core.java`. That name breaks Scala
compilation of every sibling file in `streampark-flink-shims-base` that references
`java.util`/`java.io`/`java.lang` — a nested package literally named `java` shadows the real
`java.*` standard library for any Scala file sharing the enclosing package. Reproduced both ways
in a clean build:

- `org.apache.streampark.flink.core.java` → `BUILD FAILURE`, 33 errors, all `java.util`/`java.io`/
  `java.lang` unresolved in unrelated sibling `.scala` files
- `org.apache.streampark.flink.core.javaapi` → `BUILD SUCCESS`, no other files affected

Went with `javaapi` for this PR. Open to other naming suggestions if `javaapi` isn't preferred.

## Design note: composition instead of full delegation

The legacy traits extend/mimic Flink's `TableEnvironment` / `StreamTableEnvironment` directly,
re-implementing ~40-50 delegate methods each (`FlinkStreamTableTrait` additionally needed a
`$`-prefix naming hack to avoid clashes between the two APIs). The new classes hold the
environment(s) as fields instead, exposed via `getTableEnv()` / `getEnv()`. This removes the
delegate boilerplate and the naming hack, and means the classes won't go stale as the Flink Table
API changes — but it **is a breaking change** for code that currently calls Table API methods
directly on a trait instance (e.g. `job.sqlQuery(...)` → `job.getTableEnv().sqlQuery(...)`).
Flagging for discussion since the Phase 11 migration guide will need to cover it.

## Verifying this change

```bash
mvn test -pl streampark-flink/streampark-flink-shims/streampark-flink-shims-base
```

Tests run: 9, Failures: 0, Errors: 0

## Does this pull request potentially affect one of the following parts?

- [x] The public API
- [ ] The web application
- [ ] The build system

## Documentation

- Relates to issue #4408
- Part of umbrella issue #4408, sub-issue 0.2

### DIFF
--- a/streampark-flink/streampark-flink-shims/streampark-flink-shims-base/src/main/java/org/apache/streampark/flink/core/javaapi/AbstractFlinkJob.java
+++ b/streampark-flink/streampark-flink-shims/streampark-flink-shims-base/src/main/java/org/apache/streampark/flink/core/javaapi/AbstractFlinkJob.java
@@ -1,0 +1,89 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.streampark.flink.core.javaapi;
+
+import org.apache.flink.api.common.JobExecutionResult;
+import org.apache.flink.api.java.utils.ParameterTool;
+import org.apache.flink.table.api.TableEnvironment;
+
+import java.util.function.Consumer;
+
+/**
+ * Package-private shared base for {@link FlinkTableJob} and {@link FlinkStreamTableJob} — holds
+ * the common {@code ready → handle → execute → destroy} lifecycle, the SQL bridge, and the
+ * {@code app.name} lookup, so neither subclass has to repeat them.
+ *
+ * <p>Not part of the public API: callers only ever see {@link FlinkTableJob} or {@link
+ * FlinkStreamTableJob}, never this type directly.
+ */
+abstract class AbstractFlinkJob {
+
+    protected final ParameterTool parameter;
+
+    protected final TableEnvironment tableEnv;
+
+    protected AbstractFlinkJob(ParameterTool parameter, TableEnvironment tableEnv) {
+        this.parameter = parameter;
+        this.tableEnv = tableEnv;
+    }
+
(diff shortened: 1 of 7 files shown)


wolfboys:
Thank you for your valuable contribution.
After careful evaluation, we found that the current PR introduces significant changes compared with the existing implementation and may impact the original design and migration approach. To ensure a minimal, stable, and manageable migration process, the PMC has discussed and reviewed the proposal, and we have redefined the roadmap for removing Scala.

Currently, we plan to proceed with Stage 1 — Flink Core: [4445](https://github.com/apache/streampark/issues/4445) ,  [4446](https://github.com/apache/streampark/issues/4446) ,  [4447](https://github.com/apache/streampark/issues/4447) , [4448](https://github.com/apache/streampark/issues/4448) , migrating modules incrementally according to the proposed execution plan(Principle: no logic changes; migrate modules to Java with readable, idiomatic code). We kindly suggest following the steps outlined in the issue: [flink core](https://github.com/apache/streampark/issues/4408)

Thanks again for your effort and contribution. We look forward to continuing the collaboration on this migration.

PR #4486, opened 2026-08-12
[Flink] Fix BuildPipeline.execStep NPE on steps that return null

### LINKED ISSUE #4481: [Bug] BuildPipeline.execStep throws NullPointerException for any step that legitimately returns null
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/streampark/issues?q=is%3Aissue+label%3A%22bug%22) and found no similar issues referencing `BuildPipeline`/`execStep`.

### Java Version

Temurin 21.0.11 (console), also reproduced building with Microsoft OpenJDK 11.0.28

### Scala Version

2.12.x

### StreamPark Version

3.0.0-SNAPSHOT (`dev` branch, commit `f89652b67`)

### Flink Version

2.2.1 (official binary distribution, standalone/remote cluster)

### Deploy mode

remote

### What happened

Building/releasing a Flink SQL application always fails at step 1/2 ("Create building workspace") with an uncaught `NullPointerException`, even though the workspace directory is genuinely created successfully on disk. The failure is **not visible anywhere in the normal log files** (`logs/error.*.log`) — it only shows up in the raw stdout redirect (`logs/streampark.out`), because it's thrown from a background thread pool task before the console's `GlobalExceptionHandler` is ever involved, and the persisted `ApplicationBuildPipeline.hasError`/`errorSummary`/`errorStack` fields stay `null`/`false`, making the failure look silent from the REST API (`POST /flink/pipe/detail` just shows `pipeStatus: FAILURE` with no error detail).

Root cause: `BuildPipeline#execStep()` wraps every successful step result in `Optional.of(result)`:

```java
R result = process.call();
...
return java.util.Optional.of(result);   // <-- throws NPE if result == null
```

`Optional.of(null)` always throws `NullPointerException` per the Java API contract (use `Optional.ofNullable` for that). Many pipeline steps legitimately have nothing meaningful to return — e.g. "create/clean the build workspace" just performs a side effect and returns `null` — and this is the *normal*, successful case, not an edge case. This is a classic Scala→Java migration bug (introduced by #4463, "[Migrate] Migrate streampark-flink-packer from Scala to Java"): the original Scala code almost certainly used `Option(x)`, which tolerates `null` and produces `None`, not the strict `Optional.of(x)`.

This is not specific to Flink SQL or remote/standalone deploy mode — the same `execStep(...).orElseThrow(() -> pipelineException())` pattern, with a step whose `Callable` returns `null`, appears at **~20 call sites across 6 files**: `BuildPipeline#runYarnSqlBuildSteps` (used by the Yarn pipelines), `FlinkRemoteBuildPipeline`, `FlinkK8sSessionBuildPipeline`, `AbstractK8sApplicationBuildPipeline`, `FlinkK8sApplicationBuildPipeline`, and `SparkK8sApplicationBuildPipeline`. Every deploy mode's build pipeline is affected wherever its first (or any null-returning) step runs.

Simply swapping `Optional.of` for `Optional.ofNullable` is **not sufficient** on its own: every call site relies on `.orElseThrow(() -> pipelineException())` to turn a *failed* step into a thrown exception, and `Optional<R>` cannot distinguish "empty because the step failed" from "empty because the step succeeded with a `null` result" — so a naive one-line fix just trades a crash for `.orElseThrow()` incorrectly firing on every successful null-returning step. The actual fix needs `execStep()` to signal failure via a thrown exception directly (matching what all 20 call sites already reconstruct via `orElseThrow`) rather than via `Optional` emptiness, and the corresponding call sites need to drop the now-redundant `.orElseThrow(...)`.

### Error Exception

```
06:31:23.581 [streampark-build-pipeline-executor-0] ERROR org.apache.streampark.flink.packer.pipeline.impl.FlinkRemoteBuildPipeline - [StreamPark] [streampark-packer] Building pipeline has failed. | appName=streampark_bounded_probe
java.lang.IllegalStateException: null
	at org.apache.streampark.flink.packer.pipeline.BuildPipeline.pipelineException(BuildPipeline.java:259)
	at org.apache.streampark.flink.packer.pipeline.impl.FlinkRemoteBuildPipeline.lambda$buildProcess$1(FlinkRemoteBuildPipeline.java:64)
	at java.base/java.util.Optional.orElseThrow(Optional.java:403)
	at org.apache.streampark.flink.packer.pipeline.impl.FlinkRemoteBuildPipeline.buildProcess(FlinkRemoteBuildPipeline.java:63)
	...
Caused by: java.lang.NullPointerException: null
	at java.base/java.util.Objects.requireNonNull(Objects.java:233)
	at java.base/java.util.Optional.of(Optional.java:113)
	at org.apache.streampark.flink.packer.pipeline.BuildPipeline.execStep(BuildPipeline.java:135)
	at org.apache.streampark.flink.packer.pipeline.impl.FlinkRemoteBuildPipeline.buildProcess(FlinkRemoteBuildPipeline.java:59)
	... 5 common frames omitted
```

### Code of Conduct

88fantasy:

Issue Number: close #4481

Fixes `BuildPipeline#execStep()` throwing `NullPointerException` whenever a pipeline step's `Callable` legitimately returns `null` — which is the normal case for steps that only perform a side effect (e.g. "create/clean the build workspace") and have nothing meaningful to hand to the next step. This breaks the build/release pipeline for a Flink SQL application (and, per the ~20 affected call sites, potentially every deploy mode) at its very first step.

## Brief change log

- `BuildPipeline#execStep()`: return `R` directly instead of `Optional<R>`, and throw `pipelineException()` directly from the `catch` block on failure, instead of returning `Optional.empty()` for the caller to detect via `.orElseThrow(...)`. This removes the null ambiguity entirely — Java `Optional` cannot distinguish "empty because the step failed" from "empty because it succeeded with a `null` value," which is exactly what made a one-line `Optional.of` → `Optional.ofNullable` swap insufficient (it would just make `.orElseThrow()` mis-fire on every successful null-returning step instead of crashing at `Optional.of`).
- Drop the now-redundant `.orElseThrow(() -> pipelineException())` / `.orElseThrow(this::pipelineException)` at all ~20 call sites: `BuildPipeline#runYarnSqlBuildSteps`, `FlinkRemoteBuildPipeline`, `FlinkK8sSessionBuildPipeline`, `AbstractK8sApplicationBuildPipeline`, `FlinkK8sApplicationBuildPipeline`, `SparkK8sApplicationBuildPipeline`.

No caller had extra logic in its `orElseThrow` lambda beyond `throw pipelineException();` — every occurrence checked and removed identically.

## Verifying this change

Manually verified against a real Flink 2.2.1 standalone cluster:

- Before this change: `POST /flink/pipe/build` for a Flink SQL application always fails at step 1/2 ("Create building workspace") with `NullPointerException` at `Optional.of(null)`, even though the workspace directory is genuinely created on disk. The failure is invisible via the REST API (`hasError: false`, `errorSummary: null`) and only shows up in raw stdout (`streampark.out`), not the console's own error log — because the exception is thrown from a background thread pool task before `GlobalExceptionHandler` is ever involved.
- After this change: `POST /flink/pipe/build` for the same application completes with `pipeStatus: SUCCESS` for both steps, and produces a real 6.2 MB shaded jar at `workspace/workspace/<appId>/streampark-flinkjob_<jobName>.jar`.

`./mvnw -pl streampark-flink/streampark-flink-packer,streampark-console/streampark-console-service -am clean compile checkstyle:check spotless:check` passes with 0 violations across all touched modules.

- Anything that affects deployment: no
- The persistence of application state: no
- The direction of network connections: no
- Anything that affects any api: no (internal `BuildPipeline` contract used only within `streampark-flink-packer` and its console caller; `execStep`/`runYarnSqlBuildSteps` are `protected`, not part of any public/external API)

### DIFF
--- a/streampark-flink/streampark-flink-packer/src/main/java/org/apache/streampark/flink/packer/pipeline/BuildPipeline.java
+++ b/streampark-flink/streampark-flink-packer/src/main/java/org/apache/streampark/flink/packer/pipeline/BuildPipeline.java
@@ -110,7 +110,7 @@ public abstract class BuildPipeline extends LoggerSupport
         }
     }
 
-    protected <R> java.util.Optional<R> execStep(int seq, Callable<R> process) {
+    protected <R> R execStep(int seq, Callable<R> process) {
         try {
             curStep = seq;
             stepsStatus.put(
@@ -132,7 +132,7 @@ public abstract class BuildPipeline extends LoggerSupport
                     PipelineStepStatusEnum.SUCCESS, System.currentTimeMillis()));
             logInfo("Building pipeline step[" + seq + "/" + allSteps() + "] success");
             notifyStepChange();
-            return java.util.Optional.of(result);
+            return result;
         } catch (Exception cause) {
             stepsStatus.put(
                 seq,
@@ -148,7 +148,7 @@ public abstract class BuildPipeline extends LoggerSupport
                     + "] failure => "
                     + pipeType().getSteps().get(seq));
             notifyStepChange();
-            return java.util.Optional.empty();
+            throw pipelineException();
         }
     }
 
@@ -275,8 +275,7 @@ public abstract class BuildPipeline extends LoggerSupport
                 }
                 logInfo("Recreate building workspace: " + yarnProvidedPath);
                 return null;
-            })
-                .orElseThrow(this::pipelineException);
+            });
 
         List<String> mavenJars =
             execStep(
@@ -290,8 +289,7 @@ public abstract class BuildPipeline extends LoggerSupport
                         mavenArts.stream().map(File::getAbsolutePath).collect(Collectors.toList());
                     paths.addAll(dependencyInfo.extJarLibs());
                     return paths;
-                })
-                    .orElseThrow(this::pipelineException);
+                });
 
         execStep(
             3,
@@ -301,8 +299,7 @@ public abstract class BuildPipeline extends LoggerSupport
                     YarnJarUploader.uploadJarToHdfsOrLfs(FsOperator.hdfs(), jar, yarnProvidedPath);
                 }
                 return null;
-            })
-                .orElseThrow(this::pipelineException);
+            });
     }
 
     /** intercept snapshot */
--- a/streampark-flink/streampark-flink-packer/src/main/java/org/apache/streampark/flink/packer/pipeline/impl/AbstractK8sApplicationBuildPipeline.java
+++ b/streampark-flink/streampark-flink-packer/src/main/java/org/apache/streampark/flink/packer/pipeline/impl/AbstractK8sApplicationBuildPipeline.java
@@ -65,8 +65,7 @@ abstract class AbstractK8sApplicationBuildPipeline extends BuildPipeline {
                     this::logInfo,
                     checkLocalImageFirst);
                 return null;
-            })
-                .orElseThrow(this::pipelineException);
+            });
 
         execStep(
             6,
@@ -79,8 +78,7 @@ abstract class AbstractK8sApplicationBuildPipeline extends BuildPipeline {
                     dockerProcessWatcher,
                     this::logInfo);
                 return null;
-            })
-                .orElseThrow(this::pipelineException);
+            });
 
         execStep(
             7,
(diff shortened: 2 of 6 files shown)


wolfboys (approved):
LGTM

Issue #3761, opened 2024-06-16
[Bug] Demo Flink SQL can not be run on yarn-per-job mode with flink 1.16 and flink 1.17


SbloodyS:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-streampark/issues?q=is%3Aissue+label%3A%22bug%22) and found no similar issues.

### Java Version

1.8

### Scala Version

2.12.x

### StreamPark Version

dev

### Flink Version

1.16
1.17

### deploy mode

yarn-perjob

### What happened
I have tested that shim's packaged `streampark-flinkjob_{applicationName}.jar` running properly locally using `flink run -t yarn-per-job --detached`. But it can not be run in streampark.

```
java.util.concurrent.CompletionException: java.lang.reflect.InvocationTargetException
	at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:273)
	at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:280)
	at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1606)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:750)
Caused by: java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.apache.streampark.flink.client.FlinkClient$.$anonfun$proxy$1(FlinkClient.scala:88)
	at org.apache.streampark.flink.proxy.FlinkShimsProxy$.$anonfun$proxy$1(FlinkShimsProxy.scala:62)
	at org.apache.streampark.common.util.ClassLoaderUtils$.runAsClassLoader(ClassLoaderUtils.scala:42)
	at org.apache.streampark.flink.proxy.FlinkShimsProxy$.proxy(FlinkShimsProxy.scala:62)
	at org.apache.streampark.flink.client.FlinkClient$.proxy(FlinkClient.scala:83)
[...]

PR #4496, opened 2026-08-14
[Flink] Fix Flink SQL job submission failures on the client path

### LINKED ISSUE #4483: [Bug] PipelineExecutorFactory ServiceConfigurationError when submitting a Flink job (ChildFirstClassLoader / console baseline Flink version conflict)
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/streampark/issues?q=is%3Aissue+label%3A%22bug%22) and found no similar issues referencing `PipelineExecutorFactory` or `ServiceConfigurationError`.

### Java Version

Temurin 21.0.11

### Scala Version

2.12.x

### StreamPark Version

3.0.0-SNAPSHOT (`dev` branch, commit `f89652b67`)

### Flink Version

2.2.1 (official binary distribution, standalone/remote cluster)

### Deploy mode

remote

### What happened

**This is a report of a likely root cause, not a fix** — I'm filing it separately from #4479 / #4480 / #4481 / #4482 because it appears to require touching `FlinkShimsProxy`, which this repo's own `AGENTS.md` marks as a "High-Sensitivity Area" ("Never introduce static state that could leak across classloader boundaries"), and the right fix has real architectural tradeoffs I don't think I should pick unilaterally.

After #4482's fix unblocks `FlinkClient#submit()` past its `SecurityManager` setup, job submission for a Flink SQL application against a real Flink 2.2.1 standalone cluster fails one step later with:

```
java.util.ServiceConfigurationError: org.apache.flink.core.execution.PipelineExecutorFactory:
org.apache.flink.client.deployment.executors.RemoteExecutorFactory not a subtype
```

thrown from `org.apache.flink.client.cli.GenericCLI.getExecutorFactoryNames`, called from `FlinkClientTrait.getCustomCommandLines` → `FlinkClientTrait.submit`.

**Suspected mechanism**: the console's own boot classpath bundles a concrete Flink baseline (`flink-clients-1.20.1.jar`, `flink-core-1.20.1.jar`, etc. — visible in the running process's `-classpath`), separate from the per-Flink-version jars that `FlinkShimsProxy`'s `ChildFirstClassLoader` loads for the target cluster (here, Flink 2.2.1). Both `flink-clients-1.20.1.jar` and the target cluster's `flink-clients-2.2.1.jar` declare the same `META-INF/services/org.apache.flink.core.execution.PipelineExecutorFactory` service, naming the same fully-qualified class `org.apache.flink.client.deployment.executors.RemoteExecutorFactory` in both.

`ChildFirstClassLoader#getResources()` (in `streampark-common`) collects results from **both** its own `findResources()` **and** `parent.getResources()` for every resource lookup, including `META-INF/services/*` files that `ServiceLoader` reads. That means `ServiceLoader.load(PipelineExecutorFactory.class)` running under the per-Flink-2.2.1-version classloader can end up resolving a service entry whose declaring jar and whose *actually loaded* `RemoteExecutorFactory`/`PipelineExecutorFactory` `Class` objects come from two different classloaders (the console's own `AppClassLoader`, via the 1.20.1 baseline, vs. the per-version `ChildFirstClassLoader`, via the 2.2.1 shims) — despite having identical fully-qualified names, two `Class` objects loaded by different classloaders are different types to the JVM, hence `... not a subtype`.

I have **not** verified a fix, and see at least two different directions with different tradeoffs, which is why I think this needs maintainer input rather than a PR from me:

1. Stop bundling a concrete Flink version's client jars on the console's own boot classpath (if nothing there actually needs them at that level), so `ServiceLoader` lookups under the per-version classloader only ever see one Flink version's service declarations.
2. Change `ChildFirstClassLoader#getResources()` to not merge parent results for `org.apache.flink.*` resource lookups specifically, so `ServiceLoader` running under a per-version classloader is fully isolated from whatever Flink version(s) happen to be on the console's own classpath.

### Error Exception

```
java.util.concurrent.CompletionException: java.util.ServiceConfigurationError: org.apache.flink.core.execution.PipelineExecutorFactory: org.apache.flink.client.deployment.executors.RemoteExecutorFactory not a subtype
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:315)
	...
Caused by: java.util.ServiceConfigurationError: org.apache.flink.core.execution.PipelineExecutorFactory: org.apache.flink.client.deployment.executors.RemoteExecutorFactory not a subtype
	at java.base/java.util.ServiceLoader.fail(ServiceLoader.java:593)
	at java.base/java.util.ServiceLoader$LazyClassPathLookupIterator.hasNextService(ServiceLoader.java:1244)
	...
	at org.apache.flink.client.cli.GenericCLI.getExecutorFactoryNames(GenericCLI.java:136)
	at org.apache.flink.client.cli.GenericCLI.<init>(GenericCLI.java:59)
	at org.apache.flink.client.cli.CliFrontend.loadCustomCommandLines(CliFrontend.java:1425)
	at org.apache.streampark.flink.client.trait.FlinkClientTrait.getCustomCommandLines(FlinkClientTrait.java:590)
	at org.apache.streampark.flink.client.trait.FlinkClientTrait.getCommandLineOptions(FlinkClientTrait.java:604)
	at org.apache.streampark.flink.client.trait.FlinkClientTrait.getCommandLineAndFlinkConfig(FlinkClientTrait.java:704)
	at org.apache.streampark.flink.client.trait.FlinkClientTrait.prepareConfig(FlinkClientTrait.java:325)
	at org.apache.streampark.flink.client.trait.FlinkClientTrait.submit(FlinkClientTrait.java:302)
	at org.apache.streampark.flink.client.FlinkClientEntrypoint.submit(FlinkClientEntrypoint.java:84)
	at org.apache.streampark.flink.client.FlinkClient.invokeClient(FlinkClient.java:110)
	at org.apache.streampark.flink.client.FlinkClient.submit(FlinkClient.java:64)
	...
```

- [x] Yes I am willing to submit a PR! *(diagnosis only for now — see explanation above; happy to attempt a fix once the direction is agreed)*

### Code of Conduct

### LINKED ISSUE #4488: [Bug] Build result paths are silently dropped when persisted, breaking every subsequent job start
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/streampark/issues?q=is%3Aissue+label%3A%22bug%22) and found no similar issues referencing `ShadedBuildResponse` / `buildResultJson`.

### Java Version

Temurin 21.0.11 (console), built with Microsoft OpenJDK 11.0.28

### Scala Version

2.12.x

### StreamPark Version

3.0.0-SNAPSHOT (`dev` branch, commit `9ddda84c9`)

### Flink Version

1.20.4 and 2.2.1 (official binary distributions, standalone/remote cluster)

### Deploy mode

remote

### What happened

A Flink SQL application builds successfully — `POST /flink/pipe/detail` reports `pipeStatus=3`, both steps `success`, and the fat jar really is on disk with the correct `Main-Class` — but the following `start` fails. Depending on which downstream path consumes the build result, the failure surfaces as one of three unrelated-looking errors:

- `NullPointerException` from `new File(null)` in `SubmitRequest.userJarFile()`
- `The program's entry point class 'org.apache.streampark.flink.cli.SqlClient' was not found in the jar file`
- `NullPointerException: Cannot invoke "java.io.File.getName()" because "flinkJobJar" is null`

The giveaway is in `logs/streampark.out`:

```
flinkBuildResult : { workspacePath: null, shadedJarPath: null, pass: true }
```

The build succeeded, but the object read back from the database has lost its paths.

**Root cause**: the getters on `AbstractFlinkBuildResponse`, `ShadedBuildResponse`, `K8sAppModeBuildResponse` and `DockerImageBuildResponse` do not follow JavaBean naming (`shadedJarPath()`, `workspacePath()`, `pass()`, `flinkBaseImage()`, `mainJarPath()`, `extraLibJarPaths()`, `flinkImageTag()`, `podTemplatePaths()`, `dockerInnerMainJarPath()` — no `get` prefix) and carry no `@JsonProperty`. `JacksonUtils` uses a default `ObjectMapper`, whose bean introspection does not recognise them as getters, so it **silently omits** those fields. `ApplicationBuildPipeline.onFinish()` therefore writes an incomplete JSON into `t_flink_app`'s `buildResultJson`, and `start()` deserialises that incomplete JSON back.

`pass` is the reason this went unnoticed for so long: it is dropped on write too, but its field default is already `true`, so it reads back looking correct.

This also affects the K8s Application and Docker image build responses, whose image and jar paths are persisted the same way.

Reproduced by round-tripping the response through a plain `ObjectMapper` in `jshell` against the built jar: `shadedJarPath` is absent from the serialised JSON before the fix and present after.

### Error Exception

```log
java.lang.NullPointerException: Cannot invoke "java.io.File.getName()" because "flinkJobJar" is null
	at org.apache.streampark.flink.client.impl.FlinkSessionSubmitHelper.doSubmitViaRestApi(...)
```

### Code of Conduct

### LINKED ISSUE #4489: [Bug] ClassLoaderUtils.runAsClassLoader restores a stale context classloader captured at class-init time
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/streampark/issues?q=is%3Aissue+label%3A%22bug%22) and found no similar issues referencing `ClassLoaderUtils` / `runAsClassLoader`.

### Java Version

Temurin 21.0.11 (console), built with Microsoft OpenJDK 11.0.28

### Scala Version

2.12.x

### StreamPark Version

3.0.0-SNAPSHOT (`dev` branch, commit `9ddda84c9`)

### Flink Version

1.20.4 and 2.2.1 (official binary distributions, standalone/remote cluster)

### Deploy mode

remote

### What happened

`ClassLoaderUtils.runAsClassLoader(target, supplier)` is meant to run a block under a given classloader and then put the calling thread back the way it found it. It does not: the `finally` restores `ORIGINAL_CLASS_LOADER`, a `static final` field initialised to `Thread.currentThread().getContextClassLoader()` **at class-initialisation time** — i.e. the context classloader of whichever thread first touched this class.

```java
private static final ClassLoader ORIGINAL_CLASS_LOADER = Thread.currentThread().getContextClassLoader();

public static <R> R runAsClassLoader(ClassLoader targetClassLoader, Supplier<R> supplier) {
    try {
        Thread.currentThread().setContextClassLoader(targetClassLoader);
        return supplier.get();
    } finally {
        Thread.currentThread().setContextClassLoader(ORIGINAL_CLASS_LOADER);   // <-- not the caller's
    }
}
```

The method is called from shared thread-pool threads, so for any caller whose context classloader was not that same value on entry, the "restore" silently installs a *different* classloader on that thread and leaves it there. Nothing fails at the call site; the damage lands on whatever runs on that pooled thread next.

The correct behaviour is to capture the calling thread's own value on entry and restore that. `ORIGINAL_CLASS_LOADER` is still needed by `cloneClassLoader()` and stays.

This is not a regression from the recent Scala-to-Java migration — `git log -p` shows the Scala version had the same shape.

### Code of Conduct

### LINKED ISSUE #4490: [Bug] All Flink 2.x SQL jobs fail with NoClassDefFoundError: FlinkTableInitializerV2
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/streampark/issues?q=is%3Aissue+label%3A%22bug%22) and found no similar issues referencing `FlinkTableInitializerV2` / `streampark-flink-shims-base-v2`.

### Java Version

Temurin 21.0.11 (console), built with Microsoft OpenJDK 11.0.28

### Scala Version

2.12.x

### StreamPark Version

3.0.0-SNAPSHOT (`dev` branch, commit `9ddda84c9`)

### Flink Version

2.2.1 (official binary distribution, standalone/remote cluster)

### Deploy mode

remote

### What happened

Every Flink **2.x** SQL application fails while the client builds the JobGraph:

```
NoClassDefFoundError: org/apache/streampark/flink/core/FlinkTableInitializerV2
```

thrown from `SqlClient.main()` -> `TableContext.<init>()`. Flink 1.x SQL jobs are unaffected.

`FlinkTableInitializerV2` lives in `streampark-flink-shims-base-v2`. `streampark-console-service/pom.xml` declares a normal compile dependency on the v1 module, `streampark-flink-shims-base` — so v1 ends up in the console's own `lib/` — but declares **no** equivalent dependency on `streampark-flink-shims-base-v2`. The only references to v2 are the `provided` dependencies on `streampark-flink-shims_flink-2.0/2.1/2.2/2.3`, which exist to order the reactor for `-am` builds and do not put anything into the distribution.

The result is that `TableContext` (in each `shims_flink-2.x` module) can never resolve the v2 base class it depends on at runtime.

Most likely a line missed when 2.x support was added in `e770d2e8e`.

### Error Exception

```log
java.lang.NoClassDefFoundError: org/apache/streampark/flink/core/FlinkTableInitializerV2
	at org.apache.streampark.flink.core.TableContext.<init>(...)
	at org.apache.streampark.flink.cli.SqlClient$.main(...)
```

### Code of Conduct

88fantasy:

Six defects on the path that turns a saved Flink SQL application into a submitted job. They are independent of each other, but each one only becomes reachable after the previous is fixed, which is why they are together here — each was found by submitting a real Flink SQL job to a standalone cluster and fixing whatever failed next.

Closes #4483
Closes #4488
Closes #4489
Closes #4490

## Brief change log

1. **`ClassLoaderUtils.runAsClassLoader` restores the caller's own context classloader** (#4489). It previously restored a `static final` field captured at class-initialisation time — whichever thread first loaded the class. On pooled threads that silently installs an unrelated classloader and leaves it there. `ORIGINAL_CLASS_LOADER` stays, since `cloneClassLoader()` still uses it.

2. **`FlinkClientTrait.getCustomCommandLines` and `RemoteClient.getStandAloneClusterDescriptor` run under their own class's classloader** (#4483). Both call Flink classes bound to the Flink version bundled with this module, while the calling thread's context classloader is `FlinkShimsProxy`'s target-version shims classloader. Their internal `ServiceLoader` lookups therefore resolved providers from a different Flink version than the interfaces bundled here, throwing `ServiceConfigurationError: ... not a subtype`.

3. **Build-response getters carry `@JsonProperty`** (#4488). They do not follow JavaBean naming, so Jackson silently omitted them: every persisted build result lost `shadedJarPath`/`workspacePath`, and only `pass` appeared to survive — by the coincidence that its field default is already `true`. Method names are unchanged, so no caller is affected. The K8s and Docker build responses had the same defect and are fixed with it.

4. **`SubmitRequest.userJarFile()` tolerates a null `shadedJarPath`** instead of passing it to `new File(...)`.

5. **`streampark-console-service` depends on `streampark-flink-shims-base-v2`** (#4490), so `FlinkTableInitializerV2` reaches the console's `lib/`. Only the v1 module was declared, so every Flink 2.x SQL job failed with `NoClassDefFoundError`.

6. **`setUserClassPaths` is re-enabled for `FLINK_SQL` jobs only.** It was disabled wholesale for #3761, whose reported scenario is a different deploy mode; a SQL job needs its connector jars on the client classpath. Every other job type keeps the current behaviour.

## Verifying this change

- Verified by submitting real Flink SQL jobs to a standalone Flink cluster, one defect at a time: each fix was built, deployed, and confirmed to make its specific failure disappear before the next one was investigated.
- `ShadedBuildResponse` serialisation was checked by round-tripping it through a plain `ObjectMapper` in `jshell` against the built jar — `shadedJarPath` is absent from the JSON before the change and present after.
- `mvn install` over `streampark-common`, `streampark-flink-client-api`, `streampark-flink-client-core` and `streampark-flink-packer` passes, spotless/checkstyle/scalastyle included.

- Dependencies (does it add or upgrade a dependency): **no**
- The public API: **no** (method names and signatures unchanged; only annotations and call-site classloader scoping)
- The runtime per-record code paths (performance sensitive): **no**
- Anything that affects deployment: **yes** — `streampark-flink-shims-base-v2` is now packaged into the console distribution.

## Documentation

- Does this pull request introduce a new feature? **no**

### DIFF
--- a/streampark-common/src/main/java/org/apache/streampark/common/util/ClassLoaderUtils.java
+++ b/streampark-common/src/main/java/org/apache/streampark/common/util/ClassLoaderUtils.java
@@ -39,11 +39,12 @@ public final class ClassLoaderUtils {
     }
 
     public static <R> R runAsClassLoader(ClassLoader targetClassLoader, Supplier<R> supplier) {
+        ClassLoader previousClassLoader = Thread.currentThread().getContextClassLoader();
         try {
             Thread.currentThread().setContextClassLoader(targetClassLoader);
             return supplier.get();
         } finally {
-            Thread.currentThread().setContextClassLoader(ORIGINAL_CLASS_LOADER);
+            Thread.currentThread().setContextClassLoader(previousClassLoader);
         }
     }
 
--- a/streampark-console/streampark-console-service/pom.xml
+++ b/streampark-console/streampark-console-service/pom.xml
@@ -372,6 +372,12 @@
             <version>${project.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.streampark</groupId>
+            <artifactId>streampark-flink-shims-base-v2</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
         <!-- Ensure all Flink shims are built when console-service is built with -am -->
         <dependency>
             <groupId>org.apache.streampark</groupId>
--- a/streampark-flink/streampark-flink-client/streampark-flink-client-api/src/main/java/org/apache/streampark/flink/client/bean/SubmitRequest.java
+++ b/streampark-flink/streampark-flink-client/streampark-flink-client-api/src/main/java/org/apache/streampark/flink/client/bean/SubmitRequest.java
@@ -296,7 +296,8 @@ public class SubmitRequest implements Serializable {
             } else {
                 checkBuildResult();
                 ShadedBuildResponse shadedBuildResult = buildResult.as(ShadedBuildResponse.class);
-                userJarFile = new File(shadedBuildResult.shadedJarPath());
+                String shadedJarPath = shadedBuildResult.shadedJarPath();
+                userJarFile = shadedJarPath == null ? null : new File(shadedJarPath);
             }
         }
         return userJarFile;
--- a/streampark-flink/streampark-flink-client/streampark-flink-client-core/src/main/java/org/apache/streampark/flink/client/impl/RemoteClient.java
+++ b/streampark-flink/streampark-flink-client/streampark-flink-client-core/src/main/java/org/apache/streampark/flink/client/impl/RemoteClient.java
@@ -17,6 +17,7 @@
 
 package org.apache.streampark.flink.client.impl;
 
+import org.apache.streampark.common.util.ClassLoaderUtils;
 import org.apache.streampark.flink.client.bean.CancelRequest;
 import org.apache.streampark.flink.client.bean.CancelResponse;
 import org.apache.streampark.flink.client.bean.SavepointRequestTrait;
@@ -165,13 +166,25 @@ public final class RemoteClient extends FlinkClientTrait {
 
     private Tuple2<StandaloneClusterId, StandaloneClusterDescriptor> getStandAloneClusterDescriptor(
                                                                                                     Configuration flinkConfig) {
-        DefaultClusterClientServiceLoader serviceLoader = new DefaultClusterClientServiceLoader();
-        ClusterClientFactory<StandaloneClusterId> clientFactory =
-            serviceLoader.getClusterClientFactory(flinkConfig);
-        StandaloneClusterId standaloneClusterId = clientFactory.getClusterId(flinkConfig);
-        StandaloneClusterDescriptor standaloneClusterDescriptor =
-            (StandaloneClusterDescriptor) clientFactory.createClusterDescriptor(flinkConfig);
-        return new Tuple2<>(standaloneClusterId, standaloneClusterDescriptor);
+        // DefaultClusterClientServiceLoader is bound to the Flink version bundled with this module
+        // (loaded by this class's own classloader), but the calling thread's context classloader may
+        // currently be a target-version shims classloader (see FlinkShimsProxy). Its internal
+        // ServiceLoader.load(ClusterClientFactory.class) resolves providers via the context
+        // classloader, so leaving it as the shims classloader here would load a ClusterClientFactory
+        // implementation from a different Flink version than the interface bundled here, throwing
+        // ServiceConfigurationError ("not a subtype"). Force it back to this class's own classloader
+        // for the duration of this call.
+        return ClassLoaderUtils.runAsClassLoader(
+            RemoteClient.class.getClassLoader(),
+            () -> {
+                DefaultClusterClientServiceLoader serviceLoader = new DefaultClusterClientServiceLoader();
+                ClusterClientFactory<StandaloneClusterId> clientFactory =
+                    serviceLoader.getClusterClientFactory(flinkConfig);
+                StandaloneClusterId standaloneClusterId = clientFactory.getClusterId(flinkConfig);
+                StandaloneClusterDescriptor standaloneClusterDescriptor =
+                    (StandaloneClusterDescriptor) clientFactory.createClusterDescriptor(flinkConfig);
+                return new Tuple2<>(standaloneClusterId, standaloneClusterDescriptor);
+            });
     }
 
     @FunctionalInterface
--- a/streampark-flink/streampark-flink-client/streampark-flink-client-core/src/main/java/org/apache/streampark/flink/client/trait/FlinkClientTrait.java
+++ b/streampark-flink/streampark-flink-client/streampark-flink-client-core/src/main/java/org/apache/streampark/flink/client/trait/FlinkClientTrait.java
@@ -26,6 +26,7 @@ import org.apache.streampark.common.enums.FlinkJobType;
 import org.apache.streampark.common.enums.FlinkRestoreMode;
 import org.apache.streampark.common.fs.FsOperator;
 import org.apache.streampark.common.util.AssertUtils;
+import org.apache.streampark.common.util.ClassLoaderUtils;
 import org.apache.streampark.common.util.DeflaterUtils;
 import org.apache.streampark.common.util.ExceptionUtils;
 import org.apache.streampark.common.util.FlinkConfigurationUtils;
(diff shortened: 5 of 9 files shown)


wolfboys:
Could you update to the latest branch and resolve the conflicts? I’d be happy to review it.

